PR #17, opened 2015-05-31
Add cache2k to benchmark

ben-manes:
@cruftex please review or if you send a pull request we'll drop this one

### DIFF
--- a/caffeine/build.gradle
+++ b/caffeine/build.gradle
@@ -36,6 +36,7 @@ dependencies {
 
   jmh libraries.flip_tables
   jmh benchmark_libraries.jamm
+  jmh benchmark_libraries.cache2k
   jmh benchmark_libraries.ehcache2
   jmh benchmark_libraries.ehcache3
   jmh benchmark_libraries.koloboke
--- a/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/CacheType.java
+++ b/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/CacheType.java
@@ -17,6 +17,7 @@ package com.github.benmanes.caffeine.cache;
 
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.cache2k.impl.LruCache;
 import org.cliffc.high_scale_lib.NonBlockingHashMap;
 import org.ehcache.config.Eviction.Prioritizer;
 import org.infinispan.commons.equivalence.AnyEquivalence;
@@ -24,6 +25,7 @@ import org.infinispan.commons.util.concurrent.jdk8backported.BoundedEquivalentCo
 import org.infinispan.commons.util.concurrent.jdk8backported.BoundedEquivalentConcurrentHashMapV8.Eviction;
 import org.infinispan.util.concurrent.BoundedConcurrentHashMap;
 
+import com.github.benmanes.caffeine.cache.impl.Cache2k;
 import com.github.benmanes.caffeine.cache.impl.CaffeineCache;
 import com.github.benmanes.caffeine.cache.impl.ConcurrentHashMapV7;
 import com.github.benmanes.caffeine.cache.impl.ConcurrentMapCache;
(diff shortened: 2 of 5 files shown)


cruftex:
Good job, many thanks for this!

Comments:

public V get(K key) {
  cache.get(...

I think it should be cache.peek() which is the cache2k semantic equivalent to cache.getIfPresent() in Guava. cache.get() means the cache is the authoritative source in a read-through configuration.


Cache2k_Lru and use of CacheBuilder.implementation()

In version 0.21 LRU is still the default. This will change in the next versions. CacheBuilder.implementation() is an API hazard. It will go deprecated and disappear in the near future. One of the aim in cache2k is to provide a universally good cache replacement algorithm (which is not LRU), so, consequently there should be no option to change the cache implementation on the API level. I don't want the normal user to bother with this.

If you do not use CacheBuilder.implementation() there is no need to depend on the cache2k-core module. Just the cache2k-api is enough. Everything in API module gets a proper and careful evolution.

So the question is:

Option 1: Just add a "Cache2k" to the benchmark. This will come with a replacement algorithm better than LRU in the future.

Option 2: Expose the different implementations and run benchmarks on it, like:  Cache2k_LRU, Cache2k_ClockPro or even Cache2k_Random?

I would say 1 is better. It keeps your benchmark thin and on the OTOH I don't want Cache2k to be connected with good old LRU that much any more :)

ben-manes:
> I think it should be cache.peek()

done

> One of the aim in cache2k is to provide a universally good cache replacement
> algorithm (which is not LRU), so, consequently there should be no option to change
> the cache implementation on the API level. I don't want the normal user to bother 
> with this.

This is also why Charles and I tried to keep LRU language out of the Guava api. Charles had prototypes of LIRS and ClockPro, with the LIRS code adapted by Infinispan in their version. My work showed how LRU could be made fast, and I had bad experience with other policies due to them degrading in non-obvious ways. I made attempts to augment LIRS, but being so poorly tested by the authors and subsequent implementations all having memory leaks I never trusted it enough.

For Guava we were able to adapt LRU handling for expiration and reference collection, and so being a known quantity with solid data we never revisited it. I'm not of the opinion that LRU is a bad policy, but that alternatives need more rigorous evaluation than merely reading the source material. This time around I added tracing and a primitive simulator from the start, an idea we wanted in Guava but had too much on our plate so it was backlogged indefinitely. There are many potential paths, such as [TinyLFU](http://www.cs.technion.ac.il/~gilga/TinyLFU_PDP2014.pdf) augmenting the policy rather than abandon it.

> If you do not use CacheBuilder.implementation() there is no need to depend on the > cache2k-core module. 

The core module would be a dependency, just runtime instead of compile time makes no difference for a benchmark module. I do like being explicit about what the implementation is, for now, until the library provides a canonical policy.

> Option 2: Expose the different implementations and run benchmarks on it, like: 
> Cache2k_LRU, Cache2k_ClockPro or even Cache2k_Random?

I'd go with this option for now because benchmarks need to be explicit in what they are testing, where the policy is one design tradeoff. Otherwise all libraries caches might as well be random, which can be made effectively free performance-wise.

cruftex:
BTW: I'd like to put you together a pull request. I just try to get the source into Intellij Idea 14.1, however, the needed dependencies for the JMH does not get added the the IDEA project. Probably jmh.gradle needs a tweak, but I have no idea where to start here.

ben-manes:
Most likely IntelliJ doesn't include the jmh configuration (and its sourceSet) automatically. This was the case for Eclipse, which the jmh.gradle customizes.

Its been a while since I've had to customize Gradle for IntelliJ. The [documentation](https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.idea.model.IdeaModule.html) makes me think you need to add something like,

```gradle
apply plugin: 'idea'

idea.module {
  scopes.PROVIDED.plus += [ configurations.jmh ]
}
```

PR #50, opened 2016-02-03
Add StackOverflow badge with link to tag

valery1707:


ben-manes:
I'm afraid that this might make the README too busy. It seems that people find their way to the issue tracker, email, or SO already so the hint may not be necessary.

valery1707:
Only some people's questions may be a real bug that need to be added in issue tracker. Questions asked by email is not so bad, but it not public and other people with same question will be ask it again and again.
Link to SO can show appropriate way to ask questions about project. If SO is really appropriate way :)

ben-manes:
cherry-picked onto master, thanks.

valery1707:
Thanks

PR #58, opened 2016-03-07
Adding public setters following the builder pattern to CaffeineSpec

mjburghoffer:
This will be useful for DI framework such as Guice so we can inject a strongly typed configuration object rather than a string to be parsed.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/CaffeineSpec.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/CaffeineSpec.java
@@ -23,6 +23,7 @@ import static java.util.Objects.requireNonNull;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import com.github.benmanes.caffeine.cache.Caffeine.Strength;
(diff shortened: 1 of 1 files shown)


ben-manes:
In Guice wouldn't you instead inject a `Caffeine` builder? I'd expect that to have the same effect, since it is reusable.

mjburghoffer:
Ah yes, I suppose that does make more sense.  My bad.  Thanks for looking at the PR!

ben-manes:
Appreciate the thought and PR, though! :-)

ben-manes:
hmm, actually I should clarify that the builder is reusable in regards to the `build()` method. Setting the fields like `removalListener` is not, because it can only be set once. If using setters never came up for Guava not sure if your approach or adding a `toSpec` to the builder would be helpful.

mjburghoffer:
I think for Guice, I'd just make sure to not annotate the provider with @Singleton, and i'll get a fresh one every time.  For the simple case of just setting max size, etc, it should be fine ex:

```
  @Provides
  fun provideCaffeine(): Caffeine<*, *> {
    return Caffeine.newBuilder()
        .maximumSize(2000)
        .expireAfterAccess(60L, TimeUnit.MINUTES)
  }
```

ben-manes:
Oh yeah, that makes sense. Worst case you add binding annotations to qualify them.

mjburghoffer:
I guess the one issue I can see is `Caffeine` has generic type parameters, and that can sometimes be a pain with Guice.  It might be nice to have some way to inject all the settings that are independent of the types of the keys and values of the cache.

ben-manes:
In Guice, I usually injected the built cache instead of a builder. So the spec is just an intermediate if setting from an external file. At Google that would have been a custom Flag that parsed the spec string. I'd probably do something similar with other configuration formats, like Typesafe's.

So I'd wonder if the problem is that you probably don't want to hard code the values in the Java code anyways? Then you'd need a parser for whatever mechanism you externalize the settings with, which is beyond the scope of this project to provide directly. If that's the case, simply appending to a separated string and feeding into the spec is an easy way to build a custom configuration parser.

mjburghoffer:
Realistically we'd have some sort of YAML document or app.properties for configuring the cache.  I think your first suggestion makes the most sense, just used `@Named` providers for `Caffeine`'s - or providing the cache directly.  Thanks again for the help.

PR #93, opened 2016-06-22
Demonstrate test failure for weight based evictions

drcrallen:
I came across a spurious test failure and was hoping you could clarify the expected behavior here.

The failure will happen randomly on the single shot test, but will consistently fail the repeated one.

This does not use the fancy testing annotations most of the other tests use.

### DIFF
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues/RepeatedSizeEvictionsTest.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues/RepeatedSizeEvictionsTest.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 Metamarkets Group, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.benmanes.caffeine.cache.issues;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.testing.CacheValidationListener;
+import org.testng.Assert;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+
+@Listeners(CacheValidationListener.class)
+public class RepeatedSizeEvictionsTest
+{
+  @Test
+  public void testSizeEviciton() throws Exception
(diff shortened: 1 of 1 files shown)


ben-manes:
This is expected because the eviction policy is probabilistic, not deterministic.

The exact cause is due to using a random seed in the sketch's hash function to help protect against HashDoS attacks. However that isn't strong enough when adversaries use keys with the same hash code, so there is a small amount of random admittance in the policy when a potential attack is detected. See the [design document](https://github.com/ben-manes/caffeine/wiki/Design#hashdos-protection) for the details on the HashDoS protection.

It is very likely that the policy jitter negates the need for the random seed in the hash function. I hadn't found a good reason to remove it, so I left it as a cheap security measure. I ran into the same spurious test failures though and hacked it to be a strict value in my data provider (see `RandomSeedEnforcer.ensureRandomSeed(cache)`). When that is applied your test passes.

Caffeine's policy uses a CountMinSketch to estimate the frequency of entries. This is used as a filter to avoid polluting the cache with infrequently used entries (e.g. "one-hit wonders"). Due to the small sketch size and random seed, sometimes the collisions cause `key1` to have a higher estimated frequency than `key2`, so `key2` is evicted. In larger caches this would go through an "admission window" to allow `key2` to build up its frequency to avoid consecutive misses. That's not possible due to the cache only being able to hold a single entry. The takeaway is that the eviction policy tries to exploit that a cache is inherently probabilistic, so it takes advantage of time & space efficient sketches to maximize the hit rate.

On a side note, a pending enhancement that improves the hit rate (for small recency-skewed caches) and could remove the need for the HashDoS schemes is an adaptive window. This idea is implemented in the simulator but I had the time to integrate it in yet. When added we could safely remove any randomness from the cache, though it would still be probabilistic so collisions on the key hash codes could cause similar spurious test failures if using random input.

drcrallen:
@ben-manes thanks for the excellent explanation. I can confirm setting the seed does indeed allow the test to pass, and the reason for the randomness affecting the tests makes sense.

This PR can either be closed out, or I can modify it with the thing that makes it work for example purposes if desired.

ben-manes:
Let's close. I think once I get the time to wrap up the adaptive W-TinyLFU enhancement then we should revisit the randomness to help in testability.

On a side note I was meaning to ask if you'd be interested in capturing access traces. We could then run it through the simulator to see how the eviction policy behaves with Druid's workloads and compares against other policies (e.g. LRU, OPT). It may offer insight for how to tune the cache size, though given the data volume it might not. I suspect your workloads would be similar to the [database trace](https://github.com/ben-manes/caffeine/wiki/Efficiency#database).

drcrallen:
@ben-manes I can likely do the instrumentation with the restriction that it doesn't screw around with the system more than, say, the YourKit agent, and doesn't leak confidential information that might be in the cache.

What method are you thinking of regarding capturing traces?

ben-manes:
A logger should have low overhead and you can use an async appender if concerned. The format can drop all the metadata and be just the key message (one per line). The key's `hashCode()` should be fine. Ideally we'd have a 64-bit hash, but I don't think its necessary. Then `xz` compressing the results to make it very compact. The simulator consumes long keys, has no notion of values, and handles maximum size (no weights) only. I can then give simulate ~8 sizes and chart the results so we can see the curves, and you can do the mental math to line it up with the weighted size you use.

PR #97, opened 2016-06-26
Example usage of Caffeine with a write behind cache writer.

wimdeblauwe:
Example for the write behind cache writer. It is a Maven project with the actual code and then a unit test so people can see how it can be used. It depends on the most recent releases of Caffeine and RxJava.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ jitwatch.out
 .settings
 .project
 .gradle
+.idea
+*.iml
+target
 **/hs_err_pid*.log
 **/replay_pid*.log
 **/caffeine.log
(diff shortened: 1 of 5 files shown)


ben-manes:
Thanks! This looks very nice.

I hope you don't mind that I amended your commit to make a few random clean-ups. I wanted your name displayed as the latest commit =)

P.S. I think you need to fix your git config as github is displaying your name, but not linking to your profile.

johnou:
@wimdeblauwe could you see this being used as the basis of a sync distributed cache?

wimdeblauwe:
@johnou Can you elaborate on what you mean exactly?

johnou:
@wimdeblauwe I am in search of an atomic / strongly consistent distributed or replicated (consume more memory less cost for lookup / increased redundancy?) cache for an actor system used as the activation directory.

ben-manes:
[Atomix](https://github.com/atomix/atomix) looks interesting and perhaps could be retrofitted as a cache. They [prefer](http://atomix.io/atomix/docs/comparisons/) consistency over availability, whereas the latter is more common in the various distributed caching projects.

wimdeblauwe:
I like [Hazelcast](http://hazelcast.org/) as a distributed cache. Very easy to setup, nice API.

Issue #106, opened 2016-08-07
cache performing worse than LRU


phraktle:
Hi @ben-manes,

I have created a trace from a production system where I see Caffeine underperforming a simple LRU strategy (with a cache size of 100k items, Caffeine hit rate is 72.35% vs 75.91% with LRU): [scarab-recs.trace.gz](https://github.com/ben-manes/caffeine/files/405739/scarab-recs.trace.gz)

The format is a series of longs (can be decoded with `DataInputStream#readLong`)

Any thoughts if this could be improved upon?

Thanks,
  Viktor

ben-manes:
Thanks! Did you already integrate this into the simulator or should I go ahead and do that?

phraktle:
I will submit a pull request shortly.

ben-manes:
Perfect! If we decide to check-in the trace then we should use `xz` which reduces the file from 7.1mb to 3.1mb.

ben-manes:
The results are interesting.

| Policy | Hit rate |
|---|---|
| ARC | 76.09
| Lirs | 74.30
| Lfu | 73.95
| Lru | 75.91
| Lru_TinyLfu | 67.02
| Clairvoyant | 79.25
| Unbounded | 79.25
| WindowTinyLfu | 71.57
| AdaptiveWindowTinyLfu | 72.08

It seems that frequency is a negative signal in this workload. That's surprising given how large this cache is, as usually recency has little effect past a short duration and then frequency plays a major role. The adaptive window has a maximum, so the gain is slight. The optimal window size is 100%, meaning LRU is the best we could achieve. None of the policies show an ability to improve upon LRU, which seems to be the optimum.

If we decrease the cache's size to 5k or 10k then frequency has a positive impact, increasing from `41.20` (LRU) to `46.54` (W-TinyLFU) to `47.52` (adaptive). But at a larger size its a disadvantage.

phraktle:
Interesting indeed. The workload is product recommendation lookups for several hundred e-commerce sites of varying sizes worldwide. I don't currently have an intuition as to why frequency would be a negative signal here (I would have assumed popular products should cache well based on frequency).

Any options to configure Caffeine to use an LRU policy?

ben-manes:
Not presently, as ideally users shouldn't need to worry about the implementation details. I'm hesitant to expose a setting.

I think our goal is to have a policy that the best (or very close) in all workloads. So being 3-4% worse here, while 20-30% better in other traces, is an okay trade-off.

[...]

PR #107, opened 2016-08-07
scarab trace example and parser

phraktle:
Parser and example trace related to #106.

(Code duplication with Cache2k reader, but decided it’s not worth abstracting this into a common class for now).


ben-manes:
Thanks!

PR #109, opened 2016-08-09
WIP: add JDK9 Travis job

vlsi:


build.gradle
```diff
@@ -34,7 +34,10 @@ allprojects {
 subprojects {
   apply plugin: 'com.github.ethankhall.semantic-versioning'
   apply plugin: 'nebula.provided-base'
-  apply plugin: 'net.ltgt.errorprone'
+  if (!"9".equals("$System.env.HOME")) {
```
ben-manes: Can we use [JavaVersion](https://docs.gradle.org/current/javadoc/org/gradle/api/JavaVersion.html) and keep errorprone if on JDK8?

ben-manes:
Thanks! I hadn't planned to start tackling JDK9 until its released, so I put this off. But its a good change.

build.gradle
vlsi: Thanks for the link, will try.

vlsi:
>Thanks! I hadn't planned to start tackling JDK9 until its released, so I put this off. But its a good change.

Does `caffeine` work without `Unsafe`?

I think it makes sense to start testing JDK9 to be prepared and report bugs to JDK9 team.

vlsi:
That makes sense. Afaik JDK9 beta does have VarHandles integrated.

vlsi:
Obviously akka has some problems with JDK9.

akka_2.10:2.3.15 fails with
```
   Caused by: com.sun.tools.javac.code.ClassFinder$BadClassFile: bad class file: /home/travis/.gradle/caches/modules-2/files-2.1/com.typesafe.akka/akka-actor_2.10/2.3.15/d4cdc7f139d21806a3e265c2aab51440698467dd/akka-actor_2.10-2.3.15.jar(/akka/actor/FSM$$anonfun$1.class)
      undeclared type variable: D
      Please remove or make sure it appears in the correct subdirectory of the classpath.
    	at com.sun.tools.javac.jvm.ClassReader.badClassFile(jdk.compiler@9-ea/ClassReader.java:273)
    	at com.sun.tools.javac.jvm.ClassReader.findTypeVar(jdk.compiler@9-ea/ClassReader.java:916)
    	at com.sun.tools.javac.jvm.ClassReader.sigToType(jdk.compiler@9-ea/ClassReader.java:623)
    	at com.sun.tools.javac.jvm.ClassReader.sigToTypes(jdk.compiler@9-ea/ClassReader.java:830)
    	at com.sun.tools.javac.jvm.ClassReader.classSigToType(jdk.compiler@9-ea/ClassReader.java:746)
    	at com.sun.tools.javac.jvm.ClassReader.sigToType(jdk.compiler@9-ea/ClassReader.java:659)
    	at com.sun.tools.javac.jvm.ClassReader.sigToTypeParam(jdk.compiler@9-ea/ClassReader.java:
```

and akka_2.11:2.4.9-RC2 fails with
```
aused by: com.sun.tools.javac.code.ClassFinder$BadClassFile: bad class file: /home/travis/.gradle/caches/modules-2/files-2.1/com.typesafe.akka/akka-actor_2.11/2.4.9-RC2/2983907774623d6e811d37f3e0465a78ebd6060a/akka-actor_2.11-2.4.9-RC2.jar(/akka/actor/ReflectiveDynamicAccess$$anonfun$getObjectFor$1$$anonfun$apply$1.class)
  undeclared type variable: T
```


Should I temporary disable javadoc for JDK9 somehow?

ben-manes:
Sorry I'm a little slow. Tuesdays are a bit hectic for me, so harder to steal a moment.

Disabling JavaDoc was a good idea, since few would use it directly. The few dependency-based users that I know of would prefer digging into the code anyway.

ben-manes:
I installed build 130 and checked out your branch. We should also delete `SynchronizedBenchmark` which plays with the removed `UNSAFE.monitorEnter`. Fails for a `gradle build` but JMH isn't compiled as part of an assemble.

simulator/build.gradle
```diff
@@ -32,6 +32,10 @@ test {
 
 tasks.withType(Javadoc) {
   options.addStringOption('Xdoclint:none', '-quiet')
+  if (!JavaVersion.current().isJava9Compatible()) {
```
ben-manes: I think you meant this to remove the `not` condition.

ben-manes: or maybe you meant to write it as a 1-liner.

```gradle
enabled = !JavaVersion.current().isJava9Compatible()
```

ben-manes:
Seems that Jacoco fails for tests, as reported in the [Gradle forum](https://discuss.gradle.org/t/jacoco-plugin-does-not-work-with-java-9/18157/2). Upgrading to `0.7.8-SNAPSHOT` results in,

```
    java.util.ServiceConfigurationError: sun.util.locale.provider.LocaleDataMetaInfo: Provider sun.util.resources.provider.NonBaseLocaleDataMetaInfo could not be instantiated

        Caused by:
        java.lang.NoClassDefFoundError: Could not initialize class sun.util.resources.provider.NonBaseLocaleDataMetaInfo
```

So we'd probably need to disable code coverage too.

ben-manes:
Oh, when I ran with `--no-daemon` it passed and had a test failure

```
com.github.benmanes.caffeine.OSGiTest > sanity FAILED
    java.lang.IllegalStateException: Stream handler unavailable due to: null
```

ben-manes:
and then we skip that to have mockito [bomb](https://github.com/mockito/mockito/issues/546)

ben-manes:
Seems the mockito issue was a fluke caused by Gradle retaining stale artifacts in its directories. The JDK9 build passed for me with the local changes and running,

`$ gradle test -x :caffeine:osgiTests -x :guava:test -x testCompatibilityKit`

The first two were due to OSGi, which is isolated in `:caffeine` to avoid test framework clashes. The latter is because the TCK uses `JmxMBeanServer` which is no longer available.

ben-manes:
`MBeanServerBuilder` resolves the need for `JmxMBeanServer`. That leaves only the OSGi tests.

ben-manes:
I have the primary test suite passing now.

The slow tests (`slowGuavaTest` and `slowCaffeineTest`) fails due to soft-references. It seems that JDK9 does not honor `-XX:SoftRefLRUPolicyMSPerMB=0` which is supposed to make soft references behave as weak references (predictably GC'd). This causes the `ReferenceTest` to fail.

ben-manes:
Seems the guava tests get hung on Travis but not locally. No idea why.

ben-manes:
Merged! There is some remaining issues we need to investigate. It would also be nice to have both JDK8 and JDK9 running.

PR #126, opened 2016-10-13
Update collision cache version and config.

jamespedwards42:
Finally ready to start messing around with the simulator, thanks for wiring everything up, this is very nice.  Awesome to get that immediate feedback after turning configuration knobs.

I need to rollback the changes I made to reference.conf here, but I wanted to ask, how did you run DS1 and glimpse mentioned on jamespedwards42/collision#1?  I downloaded DS1.lis from [Dharmendra S. Modha's page](http://researcher.watson.ibm.com/researcher/view_person_subpage.php?id=4700) and used the arc format, but I get less than 1% hit rates for all the caches.  Also, I couldn't figure out what trace file corresponds to glimpse.  And, is the number after the trace name the number of entries in that dataset or a simulator configuration? ... Everything worked as expected for mutli3.

I've added atomic loading by default, with an additional `getAggressive` method with the non-synchronized loading behavior.  Also, strict capacity is now the default behavior for sparse caches.

Working on updating the readme with a benchmark that shows, given a chance to sufficiently sample a dataset, caffeine will outperform everything.  Still going to have a couple of corner use cases that I think will be a good fit for collision.

Regarding the logarithmic counters for TinyLFU.  Feel free to copy paste [any of that stuff](https://github.com/jamespedwards42/collision/blob/master/src/main/java/com/fabahaba/collision/cache/AtomicLogCounters.java).  I think there might be some consistency issues with the acquire/release semantics I'm using through VarHandles though.  I need to go deeper into how that works.  The counters work as expected on my Ubuntu benchmark server, and mac os laptop with Java 9 build 138+.  But the unit test for them is consistently failing on TravisCI's Ubuntu Trusty with build 134.  Waiting for the JDK version to get updated before I look into it further.

### DIFF
--- a/simulator/src/main/resources/reference.conf
+++ b/simulator/src/main/resources/reference.conf
@@ -40,65 +40,65 @@ caffeine.simulator {
 
   policies = [
     # Policies that provide an optimal upper bound
-    "opt.Unbounded",
-    "opt.Clairvoyant",
+    #"opt.Unbounded",
+    #"opt.Clairvoyant",
 
     # Policies based on maintaining a linked-list cross-cutting the hash table
-    "linked.Lru",
-    "linked.Mru",
-    "linked.Lfu",
-    "linked.Mfu",
-    "linked.Fifo",
-    "linked.Clock",
-    "linked.S4Lru",
-    "linked.MultiQueue",
-    "linked.SegmentedLru",
+    #"linked.Lru",
+    #"linked.Mru",
+    #"linked.Lfu",
+    #"linked.Mfu",
+    #linked.Fifo",
+    #"linked.Clock",
+    #"linked.S4Lru",
+    #"linked.MultiQueue",
+    #"linked.SegmentedLru",
 
     # Policies based on obtaining a random sampling from the hash table
-    "sampled.Lru",
-    "sampled.Mru",
-    "sampled.Lfu",
-    "sampled.Mfu",
-    "sampled.Fifo",
-    "sampled.Random",
+    #"sampled.Lru",
+    #"sampled.Mru",
+    #"sampled.Lfu",
+    #"sampled.Mfu",
+    #"sampled.Fifo",
+    #"sampled.Random",
 
     # Policies based on the 2Q algorithm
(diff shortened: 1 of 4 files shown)


ben-manes:
I was thinking we might be able to merge this all into master if disabled by default. Since the jar is linked but class not loaded, it should compile and not break on JDK8. But to run with collision would require JDK9 and we can add precondition to the `CollisionPolicy` checking the system property.

Did you run DS1 with a maximum size of 5,000,000? The setting is in the config file, defaulting to a mere 512.

Glimpse is a LIRS trace named `gli.trace.gz`. The `multiX` traces come from there too and is described in the paper. I think the numbering is how many programs are being run simultaneously, or something like that.

I do think the log counters are pretty cool. At some point we should experiment with them adapted onto TinyLfu and put that into the simulator. TinyLfu is frequency biased and it would be interesting to see if log counters help balance it with recency workloads. @gilga1983 is the expert in all things TinyLFU, so he might have some good insights.

Travis has a lot more contexts switches due to other builds on the machine, virtualizing to 2 cores, more frequent GCs (less memory), etc. So it typically exposes subtle races more so than a nice laptop does, surprisingly enough. Its really frustrating to debug, though.

simulator/src/main/resources/reference.conf
```diff
@@ -40,75 +40,75 @@ caffeine.simulator {
...
     # Caching products
-    "product.Guava",
-    "product.TCache",
+    #"product.Guava",
+    #"product.TCache",
     "product.Cache2k",
     "product.Caffeine",
-    "product.Ehcache2",
+    #"product.Ehcache2",
     "product.Ehcache3",
```
ben-manes: Ehcache3 is painfully slow for large traces due to O(n) eviction (it naively scans the cache to construct a random sample). You'll want to turn it off most of the time. (It was even worse until 3.1.2 when it was MRU causing a high miss rate).

ben-manes:
I collapsed our commits with some minor clean-up, keeping you as the author, and merged.

PR #137, opened 2016-12-14
Insert into the hashtable when evicting from Am.

trauzti:
This 2Q implementation was the best I found to evaluate a C based 2Q
implementation that I have been working on.  With this change the
behavior is the same on several workloads at several data-points.

It took a lot of debugging to track this down :-)

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/two_queue/TwoQueuePolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/two_queue/TwoQueuePolicy.java
@@ -176,6 +176,7 @@ public final class TwoQueuePolicy implements Policy {
       data.remove(victim.key);
       victim.remove();
       sizeMain--;
+      data.put(node.key, node);
     }
   }
 


ben-manes:
yikes, nice find! What traces did this impact significantly?

trauzti:
I ran this on the LIRS traces and it had some effect on the hit ratio, not that much actually. I just wanted to get the exactly same hit ratio and thus I delved into this implementation to understand what was different. The logic looked the same on the surface so it was quite puzzling to see this discrepancy. To track this down I printed out the contents of A1in on every reference and compared the output to see where it started to be different and that let me to this find.

I see that you follow the 2Q pseudocode strictly on the A1in hit. i.e. "do nothing" and I agree with following the pseudocode strictly. This seems like an odd thing based on the fact from the text in the paper says that A1 is for pages seen once and Am is for pages seen more than once. Originally my implementation moved the block to Am on an A1in hit but I have now changed it to "do nothing".

ben-manes:
In the prior line, `sizeMain--`, is that still accurate? Since we're replacing I'd think the size should be the same, but you're more familiar with it at the moment.

ben-manes:
okay, nevermind, I guess that's fine.

trauzti:
Yeah that's fine. The data.put(...) could also be removed from these three cases and just put at the end of the reclaimfor function.

PR #144, opened 2017-02-18
Fix stale notification when eviction raced with an update

ben-manes:
[SOLR-10141](https://issues.apache.org/jira/browse/SOLR-10141) (thanks Yonik!)

As an optimization, an update is allowed to bypass the hash map and
synchronize on the read entry directly. In this block it checks
liveliness, perform the mutation, and notifies the writer. This avoids
more expensive computations through the map.

Previously, an eviction was performed in a computation to remove the
entry and notify the writer, or resurrect. Inside the computation the
entry was not synchronized on, and that was done only after it was
removed from the table. The removal listener was notified with the
value initially read at the start of this method.

This allowed an update to modify the value while (or after) the entry
was removing it from the hash table. This led to notifying the writer
and removal listener with the stale value. Because the writer must
be called exclusively with the mutation, this computation must use
a synchronized guard. Otherwise we might have preferred to re-read
the value when notifying the listener. This adds a slight penalty on
eviction (async) while allowing `put` to still be fast (but may block).

`putSlow` was removed as not longer needed. It was a computation-based
write that was safe from this issue. But it was only used when the new
weight was zero, as that update race would cause an incorrect eviction.
Now that the primary path is safe from this race, its unnecessary.

@yonik please review (thanks for finding this!)
@johnou (I think you mentioned that you may have observed this)


ben-manes:
Hey @johnou, you might be interested in this [slide deck](https://docs.google.com/presentation/d/1NlDxyXsUG1qlVHMl4vsUUBQfAJ2c2NsFPNPr2qymIBs/edit#slide=id.p). I didn't know your email so couldn't ping you about it, but seemed to be in your area of interest. It was for a private talk at a startup nearby a week ago.

johnou:
@ben-manes you bet I am, always welcome to reach out to me at johno.crawford@gmail.com

ben-manes:
I fixed a similar bug in when clearing the cache using`invalidateAll()` or `asMap().clear()`. I added a simpler version of @yonik's test that fails prior to the changes. I've also added unit tests that rely on explicit lock ordering of the implementation to deterministically validate the eviction and clear bugs.

I've run all the tests locally and will release after the CI confirms.

ben-manes:
Released 2.4.0

johnou:
@ben-manes did you ever hear back from Doug about the tickless hierarchical timing wheel in Java?

ben-manes:
When I brought it up he said it was on his backlog too, but I haven't talked with him about it since. I think jdk9 has a dedicated scheduler thread for the common fjp. Perhaps if its overloaded he'll add a ticking one.

I would like to add that feature by a jdk9 release, since I'll bump up the version and fix API quirks. I don't think it's hard, just requires a little motivation.

PR #150, opened 2017-04-03
Including the Rapidoid cache into the benchmarks.

nmihajlovski:


ben-manes:
Nice! 👍 
Sorry, I completely forgot that I promised to take a stab at this too.

nmihajlovski:
No worries, I was happy to contribute a bit. :) I finally found the time to do it.

ben-manes:
Do you want to add this to the Simulator as well? That shows the hit rate efficiencies.

nmihajlovski:
That would be nice, unless it's time-consuming, so I would consider it as optional.

ben-manes:
It looks like `Caching.shutdown()` isn't needed anymore, because you no longer schedule the crawler if the TTL is not set. In the benchmarks they call `cleanUp` during the setup, so it would have been turned off. But since its not loaded, I don't see any of the Rapidoid logging spamming the output. So I think I can remove that line?

ben-manes:
I amended the simulator addition to your commit and merged. Thanks!

nmihajlovski:
That's correct - `Caching.shutdown()` isn't needed if TTL isn't set. Thanks for the help!

PR #152, opened 2017-04-26
fix OHC gradle dependencies

ohadeytan:
the `window-tinylfu-SNAPSHOT` no longer existed/needed


ben-manes:
Thanks! I had fixed this in my [expiry branch](https://github.com/ben-manes/caffeine/tree/expiry), but I'm still working on unit tests so hadn't merged it in yet.

PR #162, opened 2017-06-05
Refactored Function<Integer,Boolean> to IntPredicate

Harshrajsinh:
Hello,

I am a graduate student at Oregon State University and as a part of my research and subject CS562 Applied Software Engineering project (Study and Refactoring of Functional Interface in Java), I have done refactoring of the Function<Integer,Boolean> to IntPredicate.  My project deals with the boxing and unboxing of Wrapper class and the primitive data-types. I want to know that whether the open source community is ready to accept these micro-optimizing refactorings or not. 

Thank you,
Harsh Thakor

### DIFF
--- a/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/ComputeBenchmark.java
+++ b/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/ComputeBenchmark.java
@@ -20,6 +20,7 @@ import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


ben-manes:
Thank you, but I think this actually introduces boxing and the benchmark was written specifically to avoid that overhead.

The key space is a precomputed array of `Integer[]` objects, with the selected object passed through the function to the cache's method. The caches are all object based, so a primitive would have to be boxed. The entry is always present and the value returned all the way back to the harness so that the JIT doesn't dead code eliminate the calls.

Your change introduces unboxing the `Integer[]`, which would be trivial to fix. Unfortunately since the caches all require object keys there is a necessary boxing within the function. The only way to avoid this is to use a boxed type throughout.

PR #170, opened 2017-07-03
Fix IllegalStateException being thrown for all event types.

nickrobison:
I ran into a problem where an IllegalStateException was being thrown every time my event listeners were invoked.

```
Jul 02, 2017 9:30:20 PM com.github.benmanes.caffeine.jcache.event.EventTypeAwareListener dispatch
WARNING: null
java.lang.IllegalStateException: Unknown event type: REMOVED
	at com.github.benmanes.caffeine.jcache.event.EventTypeAwareListener.dispatch(EventTypeAwareListener.java:80)
	at com.github.benmanes.caffeine.jcache.event.EventDispatcher.lambda$publish$2(EventDispatcher.java:215)
	at java.util.concurrent.CompletableFuture.uniRun(CompletableFuture.java:705)
	at java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:687)
	at java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:443)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

I traced it down to what appears to be a missing default case in the EventType switch statement. This patch resolves the issue for me and passes the relevant tests.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/event/EventTypeAwareListener.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/event/EventTypeAwareListener.java
@@ -66,16 +66,16 @@ final class EventTypeAwareListener<K, V> implements CacheEntryCreatedListener<K,
       switch (event.getEventType()) {
         case CREATED:
           onCreated(event);
-          break;
+          return;
         case UPDATED:
           onUpdated(event);
-          break;
+          return;
         case REMOVED:
           onRemoved(event);
-          break;
+          return;
         case EXPIRED:
           onExpired(event);
-          break;
+          return;
       }
       throw new IllegalStateException("Unknown event type: " + event.getEventType());
     } catch (Exception e) {


ben-manes:
Thanks! It looks like a mistake when I made changes to resolve ErrorProne warnings. For some reason they disliked `default` and wanted the switch to break out. I must have not noticed that had `break` instead of `return` like other cases. Does ErrorProne generate warnings for that still?

ben-manes:
I think the error was due to [UnnecessaryDefaultInEnumSwitch](http://errorprone.info/bugpattern/UnnecessaryDefaultInEnumSwitch).

nickrobison:
I think in general it still checks for that, not seeing any warnings during build.

Perhaps a better solution would be to return from each case statement?

Maaartinus:
> Perhaps a better solution would be to return from each case statement?

Agreed. An enum-switch without a default clause is more future-proof as you'll get a warning when a new enum member gets created. That's actually the reason for `UnnecessaryDefaultInEnumSwitch`.

Shouldn't the `IllegalStateException` be an `IllegalArgumentException` (it does depend on the input)?
It gets thrown and immediately caught, that's a bit strange, but tuning something what can't really happen doesn't make much sense.

ben-manes:
I think you guys are right. Please switch to return instead of break and drop the default, which is how I should to have fixed this originally. If you'd like to change to `IllegalArgumentException` then we should do a quick scan to make usages consistent.

PR #182, opened 2017-08-31
Allow unwrapping of a JCache cache to a native Caffeine cache

kdombeck:
We are using Caffeine with JCache and need access to the native Caffeine cache to register it with [Prometheus](https://github.com/prometheus/client_java/blob/master/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java). The current implementation just returns itself.

This is very similar to the way [Ehcache](https://github.com/ehcache/ehcache-jcache/blob/598c609d3f1d5addcdb866f64b37b9769d2fd97b/ehcache-jcache/src/main/java/org/ehcache/jcache/JCache.java#L726-L732) is doing it.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
@@ -887,6 +887,9 @@ public class CacheProxy<K, V> implements Cache<K, V> {
 
   @Override
   public <T> T unwrap(Class<T> clazz) {
+    if (clazz.isAssignableFrom(cache.getClass())) {
+      return clazz.cast(cache);
+    }
     if (clazz.isAssignableFrom(getClass())) {
       return clazz.cast(this);
     }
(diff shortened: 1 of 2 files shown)


ben-manes:
That seems reasonable. Thanks for adding the test case, too.

Let me know if you want me to kick off a release.

kdombeck:
Thanks for the quick turnaround. After looking into this further it looks like if we use Caffeine with JCache we can only get JCache JMX stats and not the native Caffeine stats, so I won't be able to register this with Prometheus.

kdombeck:
It was the [load](https://github.com/prometheus/client_java/blob/master/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java#L45-L48) times that would be really useful. Plus the times in [JCache](https://static.javadoc.io/javax.cache/cache-api/1.0.0/javax/cache/management/CacheStatisticsMXBean.html) appear to be kinda useless from a Prometheus standpoint since it returns the average rather than the raw total time.

I am going to work on a generic JCache module for Prometheus and create a PR if it looks good. 

Thanks for your help.

ben-manes:
I think also we'd need to have a flag to turn on `Caffeine.recordStats()` if you were to rely on unwrapping. Perhaps the internal stats could be proxied, but I think there were some minor distinctions due to expiration. The adapter was written prior to having native variable expiration, which added a lot of bridging complexity (I disliked the lazy or O(lg n) approaches; harder work but got O(1) via timer wheel in native). I plan on revamping the adapters in v3 (targeting jdk9) to hopefully simplify it, but that requires changing the configuration and therefore delayed to a major version bump.

kdombeck:
Makes complete sense. Again thank you

PR #194, opened 2017-10-25
Allow configuration of max size and expiry simultaneously for JCache

kdombeck:
The ```eager-expiration``` will not take affect with the following configuration 
```
test-cache {
  policy {
    eager-expiration {
      after-write = 10m
    }

    maximum {
      size = 1000
    }
  }
}
```
As you can see by this [code](https://github.com/ben-manes/caffeine/blob/master/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheFactory.java#L159-L161).
```
      boolean evicts = configureMaximumSize() || configureMaximumWeight()
          || configureExpireAfterWrite() || configureExpireAfterAccess()
          || configureExpireVariably();
```
The first configuration that it finds is the only one that gets configured.

I looked at passing mocks into ```CacheFactory``` but doesn't seem possible since the configuration gets cloned. So please let me know if there is a better way I should be testing this.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheFactory.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheFactory.java
@@ -156,9 +156,9 @@ final class CacheFactory {
 
     /** Creates a configured cache. */
     public CacheProxy<K, V> build() {
-      boolean evicts = configureMaximumSize() || configureMaximumWeight()
-          || configureExpireAfterWrite() || configureExpireAfterAccess()
-          || configureExpireVariably();
+      boolean evicts = configureMaximumSize() | configureMaximumWeight()
+          | configureExpireAfterWrite() | configureExpireAfterAccess()
+          | configureExpireVariably();
       if (evicts) {
         configureEvictionListener();
       }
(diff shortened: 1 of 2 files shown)


ben-manes:
Wow, I feel pretty foolish.

I think your test looks good. I don’t recall why the configuration is cloned, but it might be for consistency of a different instance, e.g. if reusing the configuration to create multiple caches.

I’ve been swamped and trying to find the time to catch up. I have another fix promised but not sure if I can get to it soon. Is there a timeframe you need this released by?

kdombeck:
No timeframe on my side. Thank you for being so responsive with issues. Truly you are the most responsive of any open source project I have ever worked on. 

Thanks again.

ben-manes:
Great. I'll try to catch up over the weekend, so I'll leave this open to prod me.

If you have any concerns or improvements on the JCache implementation, please let me know. Admittedly it doesn't get as much attention. There are upcoming changes in [JCache 1.1](https://jcp.org/aboutJava/communityprocess/maintenance/jsr107/107MR1-summary.html) and I have the revised TCK mostly passing locally. I also plan on an overhaul for 3.0, when I get the time to start on it, so we could make braking changes if you have suggestions.

kdombeck:
The only thing I can think of is enabling the native Caffeine statistics. The stats in Caffeine are a lot better than the ones exposed in JCache.

ben-manes:
Released. Thanks again for PR!

PR #196, opened 2017-10-31
Added a note about camel-caffeine

oscerd:
Just a little note about the camel-caffeine component in Camel.

https://github.com/apache/camel/tree/master/components/camel-caffeine

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,12 +49,13 @@ Use Caffeine in a community provided integration:
  * [ScalaCache][scala-cache]: Simple caching in Scala
  * [jooby][jooby]: Modular micro framework
  * [Druid][druid]: Real-time analytics
+ * [Camel-caffeine][camel-caffeine]: In Camel 2.20.0 a specific Camel component for Caffeine has been released
 
 Powering infrastructure near you:
  * [Cassandra][cassandra]:  Manage massive amounts of data, fast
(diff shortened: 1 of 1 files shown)


ben-manes:
Maybe we should have Camel in both sections? One is for projects using it internally, the other is for integrations for consumers to take advantage of. Camel is in both. Or perhaps we should include Camel only in the latter since that's more meaningful to users? If you render the README, you'll see its kind of a mess with this change due to how verbose the combined phrasing is.

I guess I'd favor either leaving it as is, or moving Camel into the other section and linking to the documentation of `camel-caffeine` rather than the project. Thoughts?

oscerd:
I guess the origin position for Camel is good as is, maybe adding a note about camel-caffeine somewhere in the docs can be useful too.

ben-manes:
How about I we move it to the other section and change the link to,
https://github.com/apache/camel/blob/master/components/camel-caffeine/src/main/docs/caffeine-cache-component.adoc

Presumably that's useful for users and we should be more interesting in advertising that extension than an implementation detail.

oscerd:
Do you mean something like this?

ben-manes:
I made a tad simpler change, if that's okay.

oscerd:
Great thanks :-)

Issue #110, opened 2016-08-09
Notable part of caffeine.jar is consumed by generated classes


vlsi:
caffeine-2.3.1.jar -- 968137 bytes
caffeine-2.3.1.jar without LocalCacheFactory$... and NodeFactory$... classes -- 248414 bytes

Are there any ideas/plans to reduce that?

ben-manes:
The intent is of course to reduce the in-memory overhead due to the many configuration options at the cost of larger disk footprint (unloaded classes).

I think the large enum used by `NodeFactory` could be removed. The simplest would be to `getFactory` compute a numeric code, and return a new factory instance whose methods switch on that code. This would remove the bloat for a tiny runtime cost of selection.

`LocalCacheFactory` has a string switch statement. Using a numeric code would be slightly smaller by reducing the interned strings.

I tried to use inheritance to reduce duplicate code. I think Node was optimal and LocalCache was close, but not perfect. Perhaps there are some more tricks to optimize reuse, e.g. interfaces with default methods. We could also remove the modifiers, as perhaps the JVM is creating unnecessary bridge methods.

The biggest win would be to review `LocalCacheFactory` features to determine if they carry their weight. Perhaps that factory should be removed and all the fields present, but null'd if unused. The top-level cost is smaller than per-node, which was my original intent for using codegen.

Maaartinus:
> The biggest win would be to review LocalCacheFactory features to determine if they carry their weight. Perhaps that factory should be removed and all the fields present, but null'd if unused. The top-level cost is smaller than per-node, which was my original intent for using codegen.

I could imagine, that keeping the small factories small makes sense, but for example, `WISMW` has tons of fields. Adding all subclass fields to it would only grow it by a few percent. So you could generate only `WISMWAWR` (with some runtime tests) and use it in place of `WISMW`  and all its subclasses. I might be using wrong names as without being really familiar with the code, they're hard to tell apart.

But such optimizations are ugly. For it, to be efficient, you may have to reorder the hierarchy so that the heaviest additions come first. Let's say, you deal with `evicts` first. Any evicting factory is already huge and you can add all fields and save half the hierarchy. Adapting the generator could be rather easy (nice code!), but changing the hierarchy could be confusing and the reward unsure.

[...]

PR #197, opened 2017-11-01
Compile generated classes withoug debug info

jvassev:
Addressing https://github.com/ben-manes/caffeine/issues/110.
Just by not emitting debug symbols for the generated classes 200k can be saved.
The price to pay is recompiling the generated classes twice.

```
object       | before    | after     | savings (bytes)
------------------------------------------------------
class files  | 2,431,199 | 1,793,522 | 637677
caffeine.jar |   816,949 | 1,009,405 | 192456
```

### DIFF
--- a/caffeine/build.gradle
+++ b/caffeine/build.gradle
@@ -8,6 +8,9 @@ sourceSets {
   main {
     java.srcDir "${buildDir}/generated-sources/"
   }
+  gen {
+    java.srcDir "${buildDir}/generated-sources/"
+  }
 }
 
 configurations {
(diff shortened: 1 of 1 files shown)


ben-manes:
Oh, that is an clever hack! Thank you!

caffeine/build.gradle
```diff
@@ -55,6 +58,14 @@ dependencies {
   javaPoetCompile libraries.commonsLang3
 }
 
+task compileNoDebug(type: JavaCompile) {
+  source = sourceSets.gen.java.srcDirs
+  destinationDir = compileJava.destinationDir
+  options.debug = false
+  options.incremental = false
```
ben-manes, line 65: I'm reworking this so that the compilation process doesn't overwrite the files (causes cache busting, harder to detect malicious supply chain attacks). By chance do you remember why `incremental` was false?

PR #199, opened 2017-11-04
Make a Node a factory for itself

jvassev:
How about this: make the node class a factory for itself :) (details in the commit message)

Zero reflection on the fast path + 90KB classes less.

It is not how a real gentleman would structure their Java code but the circumstances are extreme :)

### DIFF
--- a/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/NodeFactoryGenerator.java
+++ b/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/NodeFactoryGenerator.java
@@ -15,42 +15,10 @@
  */
 package com.github.benmanes.caffeine.cache;
 
-import static com.github.benmanes.caffeine.cache.Specifications.BUILDER_PARAM;
-import static com.github.benmanes.caffeine.cache.Specifications.DEAD_STRONG_KEY;
-import static com.github.benmanes.caffeine.cache.Specifications.DEAD_WEAK_KEY;
-import static com.github.benmanes.caffeine.cache.Specifications.PACKAGE_NAME;
-import static com.github.benmanes.caffeine.cache.Specifications.RETIRED_STRONG_KEY;
-import static com.github.benmanes.caffeine.cache.Specifications.RETIRED_WEAK_KEY;
-import static com.github.benmanes.caffeine.cache.Specifications.kRefQueueType;
-import static com.github.benmanes.caffeine.cache.Specifications.kTypeVar;
-import static com.github.benmanes.caffeine.cache.Specifications.keyRefQueueSpec;
-import static com.github.benmanes.caffeine.cache.Specifications.keyRefSpec;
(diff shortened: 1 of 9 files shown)


ben-manes:
I guess the only flaw is that this adds extra instance fields due to the factory node's unused ones? The alternatives would all have to add some, too, of course.

jvassev:
Not really, the factory is stateless. Take the generated `WSo` for example:
```java
    ...:
   // factory methods
    public <K2, V2> Node<K2, V2> newNode(K2 key, ReferenceQueue<K2> keyReferenceQueue, V2 value,
        ReferenceQueue<V2> valueReferenceQueue, int weight, long now) {
      return new WSo<>(key, keyReferenceQueue, value, valueReferenceQueue, weight, now);
    }

    public <K2, V2> Node<K2, V2> newNode(Object keyReference, V2 value,
        ReferenceQueue<V2> valueReferenceQueue, int weight, long now) {
      return new WSo<>(keyReference, value, valueReferenceQueue, weight, now);
    }

    public <K2> Object newLookupKey(K2 key) {
      return new LookupKeyReference<K2>(key);
    }

    public <K2> Object newReferenceKey(K2 key, ReferenceQueue<K2> referenceQueue) {
      return new WeakKeyReference<K2>(key, referenceQueue);
    }
...
```

ben-manes:
> There are no instance fields in the factory as it is now an interface.

Sure there is. `WSo` has a key and value field.

jvassev:
Yes, the Wso used as factory will have some extra fields, but the Wso instances used as `Node` will have 4 more methods - no extra fields.

ben-manes:
Yep, that's all I meant. It wasn't meant as more than an observation

jvassev:
Sorry, I thought you are concerned with the bloating the Node footprint :)

jvassev:
Btw the K2, V2 are there to fix a few warnings. If we can live with the shadowing warnings that's 2K more saved and the psychological barrier of 700KB is passed :)

ben-manes:
That would be okay or you could make `NodeFactory` typed. Then no shadowing would occur.

I'm not excited at the extra top-level fields, but I am also trying to see if there is any justification for that feeling. Its probably the right tradeoff.

ben-manes:
When looking at the method handle approach, I guess it actually is worse because each `MethodHandle` holds a bunch of metadata fields. That prototype is at 680kb, with more work needed, but in comparison I think yours is nicer.

jvassev:
The enum constants would also create 144 field + 144 classes (in the metaspace) when probably 2-3 are really used. I would expect that memory usage is reduced a bit - assuming an app has a few caches now it would also need a create a few nodeFactories. The enum approach would always create 144 factories.

ben-manes:
Oh I am definitely for removing the enum. It was just easy to write when trying to get the codegen working. There was a lot of upfront development costs to get this project going.

I can merge this unless you are still iterating on any of it?

ben-manes:
I typed the `NodeFactory` and its now at 700kb. Thanks! :)

PR #200, opened 2017-11-06
Generate top-level classes for Nodes and Factories

jvassev:
If nested, every method/class/field descriptor carries
the enclosing's class name and a $ sign.
Also slightly shorter name using french words to avoid name
collisions.


ben-manes:
A few minor concerns.
 - All the classes are public, so we should make it generate package scoped.
 - The suppress warnings no longer applies. Would adding the annotation everywhere add back the bloat?

ben-manes:
I changed the visibility and added the `SuppressWarnings` annotations. Since  the retention policy is `SOURCE` it doesn't end up in the byte code.

Thanks again :)

jvassev:
Hey, np. I was about to do the same :)

From this point I think only using a minifier/obfuscator can help.

PR #202, opened 2017-11-12
Exclude stats, removalListener, expirationTicker from codegen

jvassev:
Basic implementation is added in the BoundedLocalCache class and
declared final.
Generated classes take care only of methods that throw
UnsupportedOperationException.

At the cost of 5 fields in the base class, jar size drops to 430K.


ben-manes:
I think any final methods that use the field should be inlined directly.

I'm surprised expirationTicker is on the list, since the other expiration types have their own fields. I guess that's only due to making `Expiry` a field? I'd wonder if there was a better candidate since `Expiry` is very unlikely to be used, as after-write and after-read are based on a LinkedDeque. But then `timerWheel` is still generated, I presume, so it seems there's something here to review.

I'm kind of swamped and burned out right now, so I might not get to this immediately. Thanks for this, though! :)

ben-manes:
I'm sorry that we never worked through this. In retrospect, I think you are probably right at the 5 fields being a good compromise for a smaller jar. At the time I was hesitant as this is a server-side dependency, so the disk size is less important for unloaded classes (vs runtime). However there is a lot of shading and now AOT via graal. Plus 4 of those fields are fairly commonly used features.

Do you want to rebase it? Otherwise I might try later on to see if I can redo this work.

jvassev:
I haven't had too much `caffeine` in a while...
Probably it is safer for you to complete this :)

ben-manes:
Okay, thanks! I'll close this and review the PR when slimming down the jar.

Issue #187, opened 2017-09-19
Seems like a bug...


chrisstockton:
I'm doing an update on a cached entry using an ```EntryProcessor```.  The first time that I call ```cache.invoke()```, the non-existing value is inserted into the cache and a second call to ```cache.invoke()``` does not trigger a ```loader.load()```.

Once the cached entry is expired however, any subsequent calls to ```cache.invoke()``` always trigger a call to ```loader.load()```.

Here's a test case:

```
  @Test
  public void itLoadsTwice() throws InterruptedException {
    CachingProvider provider = Caching.getCachingProvider(CaffeineCachingProvider.class.getName());
    CacheManager manager = provider.getCacheManager(provider.getDefaultURI(), provider.getDefaultClassLoader());
    Cache<String, Integer> cache = manager.getCache("test-cache");
    cache.invoke("a", new EntryProcessorImpl());
    assertEquals(1, CacheLoaderImpl.loads);

    cache.invoke("a", new EntryProcessorImpl());
    assertEquals(1, CacheLoaderImpl.loads);

    // Expire the entry
    Thread.sleep(5000);

    // expect only 1 more load here, but there were 2
    cache.invoke("a", new EntryProcessorImpl());
    assertEquals(2, CacheLoaderImpl.loads);

    // don't expect a load here
    cache.invoke("a", new EntryProcessorImpl());
    // fails, CacheLoaderImpl.loads == 3 here
    assertEquals(2, CacheLoaderImpl.loads);
  }

  static class EntryProcessorImpl implements EntryProcessor<String, Integer, Object> {
    @Override
    public Object process(MutableEntry<String, Integer> entry, Object... arguments) throws EntryProcessorException {
      Integer value = entry.getValue();
      if (value == null) {
        value = 0;
      }
      entry.setValue(++value);

      return null;
    }
  }

  static Map<String, Integer> map = new HashMap<>();

  public static class CacheWriterImpl implements CacheWriter<String, Integer> {
    static int writes = 0;

    @Override
    public void write(Entry<? extends String, ? extends Integer> entry) throws CacheWriterException {
      writes++;
      map.put(entry.getKey(), entry.getValue());
    }

    @Override
    public void writeAll(Collection<Entry<? extends String, ? extends Integer>> entries) throws CacheWriterException {
      entries.forEach(this::write);
    }

    @Override
    public void delete(Object key) throws CacheWriterException {
      map.remove(key);
    }

    @Override
[...]

PR #210, opened 2017-12-12
JCache Cache.invoke does not update entry after initial load

kdombeck:
If the cache is empty and you call ```jcache.invoke(key, entryProcessor)``` it will load the cache with the cache loader but then it will not write the updated entry with the cache writer. The reason is because the ```EntryProcessorEntry.action``` is LOADED rather than UPDATED.

Possibly related to https://github.com/ben-manes/caffeine/issues/187.

This is currently just a unit test showing the problem, it does not contain a fix. I'm not sure which direction to go for a possible fix.

### DIFF
--- a/jcache/src/test/java/com/github/benmanes/caffeine/jcache/processor/EntryProcessorTest.java
+++ b/jcache/src/test/java/com/github/benmanes/caffeine/jcache/processor/EntryProcessorTest.java
@@ -50,11 +50,13 @@ public final class EntryProcessorTest extends AbstractJCacheTest {
   private final Map<Integer, Integer> map = new HashMap<>();
 
   private int loads;
+  private int writes;
 
   @BeforeMethod
   public void beforeMethod() {
     map.clear();
     loads = 0;
+    writes = 0;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


ben-manes:
Thanks for the test, again! :)

I'm swamped today, but I'll try to look at this on my way home (PST). I owe people a new release so I should be able to get this out soon. Not sure about the fix yet until we both dig in a bit.

ben-manes:
hmm.. you are right that this is confusing. I originally followed the [reference implementation](https://github.com/jsr107/RI/blob/master/cache-ri-impl/src/main/java/org/jsr107/ri/processor/EntryProcessorEntry.java), but the JSR authors do not consider it good quality. The [specification](https://docs.google.com/document/d/1MZQstO9GJo_MUMy5iD5sxCrstunnQ1f85ekCng8LcqM/edit#heading=h.qrqsdasu42ih) is primarily the interface JavaDoc pasted into sections rather than additional depth.

The right thing is to probably sketch out the scenarios from scratch, using the examples in `EntryProcessor's` JavaDoc as guidance. I think part of the confusion is that it states "Only the net effect of multiple operations has visibility outside of the Entry Processor", but it also states weird things like `The first get and the second put only are visible to the ExpiryPolicy` which means it isn't entirely flattened out. So I don't know if we can rely on a single state machine, should use a mixture of flags, or a queue of events.

ben-manes:
The fix passes your test. The lifecycle is still very confusing, so please let me know if you discover other bugs. Adding the simple check of whether the entry was originally present in the cache seems to pass all the tests.

ben-manes:
Released

kdombeck:
Works perfectly!! Thank you

PR #238, opened 2018-05-01
Specify when and how StatsCounter#recordLoadSuccess and recordLoadFailure events are recorded more precisely

### LINKED ISSUE #237: Make the specification of StatsCounter.recordLoadFailure() more precise
Currently the specification of `StatsCounter.recordLoadFailure()` says:

> Records the failed load of a new entry. This should be called when a cache request causes an entry to be loaded, but either no value is found or an exception is thrown while loading the entry. In contrast to `{@link #recordMisses}`, this method should only be called by the loading thread.

First, it should be clearer if "loading" comes only to the loading procedures provided during the cache building, or also when `Cache.get(k, mappingFunction)` is called. Secondly, the specification of `StatsCounter.recordLoadFailure()` doesn't mention that when `Cache.get()` is called and the mapping function returns null, it is also considered a failure, it only mentions exception throwing.

leventov:
Fixes #237


ben-manes:
Thanks! I tweaked the wording to be a little more broad on `StatsCounter` and decided to not include it on the `Cache` interfaces. My concern is that a user might have to think through a lot of cases, with stats being the least important, and it may be a bit overwhelming.

PR #263, opened 2018-08-23
Enable configuring the window cache size via policy/eviction

blsemo:
Adding a low-level method to the `Eviction` interface that allows resetting the window/main percentage on an already created cache. The default behaviour/builder stays exactly the same, but it will be possible to reconfigure the cache ratio.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
@@ -453,6 +453,8 @@ abstract class BoundedLocalCache<K, V> extends BLCHeader.DrainStatusRef<K, V>
   @GuardedBy("evictionLock")
   void setMaximum(long maximum, double mainPercentage) {
     requireArgument(maximum >= 0);
+    requireArgument(mainPercentage >= 0);
+    requireArgument(mainPercentage <= 1.0d);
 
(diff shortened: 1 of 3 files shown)


blsemo:
I haven't signed the CLA yet - this PR might have to wait until I get the all clear from our management/legal.

ben-manes:
Thanks @blsemo! I'll migrate `TestEdenResize` to using the parameterized infrastructure.

blsemo:
Hi Ben,
Those adaptive tests look good - they might make the manual window size configuration unneeded. I have raised https://r3-cev.atlassian.net/browse/CORDA-2562 for us to give the branch a try and double check that it delivers the performance we need in our system. If that is the case, this feature would be very good to have in an official caffeine release.
Not sure I fully understood your last question about hiding the detail? What detail do you want to hide from whom?

ben-manes:
Thank you.

Sorry for the confusion, the last question was meant to be a repeat of the former.

I need to write unit tests still. I hope to make progress towards having a release soonish.

ben-manes:
This work is now in master! 😺

ben-manes:
Here is a mixed trace of Corda large, 3 loops, and Corda large. Caffeine achieves a hit rate of 39.5% where the optimal is 40.3%, and all other policies are at 20% or lower. I am really happy with these results, but would love your confirmation when you have the time 😄 

mixed

PR #280, opened 2018-11-11
Fix typos

imasahiro:


ben-manes:
Thanks! How did you find all of those?

PR #307, opened 2019-03-11
fix a wrong argument check while configure recordStats by CaffeineSpec

justgame:


ben-manes:
Are you sure? This asserts that there is no value par (e.g. `recordStats=true`). The condition that `value == null` is asserted. See `CaffeineSpecTest` for the unit test.

justgame:
I'm not so sure. I just configure caffeine in a springboot project, and use `spring.cache.caffeine.spec=recordStats=true,maximumSize=5000,expireAfterAccess=600s`, then it throws a   IllegalArgumentException with message "record stats does not take a value"

ben-manes:
Right, the stats (as originally defined by Guava) would be

```
spring.cache.caffeine.spec=recordStats,maximumSize=5000,expireAfterAccess=600s
```

justgame:
Sorry, it's my fault. Thank you for your explanation.

ben-manes:
No problem, it is a little confusing. Maybe we could make it optionally handle true/false in the future.

Issue #309, opened 2019-03-15
Batching of AsyncCacheLoader calls


sheepdreamofandroids:
I'm using an async cache with an AsyncCacheLoader that delays loading until a certain period has elapsed or a maximum number of keys is collected. After that delay a bulk call is done because that is way more efficient than many single calls.
Would you be interested in a contribution of a generic version of this AsyncCacheLoader implementation? Or maybe a different approach to this problem?

ben-manes:
Thanks, that makes sense. I would accept it in the `examples` directory and then we might consider if it should be promoted as a supported feature. That way others can benefit from your work without library design (e.g. scope) being a blocking consideration.

I do wonder if ReactiveX / Reactor would make this easy to implement, like it did in `write-behind-rxjava` for collecting writes to batch update? If so then it might not be much code and simpler as an example integration between libraries.

sheepdreamofandroids:
Actually I did an implementation using rxjava and the buffer() operator but I'm not completely happy about it.
Firstly the buffer() operator produces useless empty batches. They can easily be ignored but it just doesn't feel right.
Secondly it's quite easy to miss completing some futures which makes the whole thing less robust than I like. This is the main reason I like to abstract it out.
Thirdly, after a batch of keys has been processed, the times shouldn't start right away, but only when the first key is received to avoid non-optimal small batches.

ben-manes:
That makes sense and I agree those are the characteristics you'd expect. I haven't gotten used to the reactive-x libraries yet due to poor fits in my day-to-day work. Do you want to submit your implementation as an example?

sheepdreamofandroids:
Sure, just have to get around to it.
I wasted some time trying to get the project to build; the javadoc task just wouldn't work. I'll just comment it out for now.

ben-manes:
The project is built for JDK8 presently, though I suppose we could fix the Javadoc to conform to the later JDKs.

The `examples` are self-contained builds so you shouldn't be impacted if you go that route.

ben-manes:
I'll close this and we can discuss again if you send a PR. 😄

PR #316, opened 2019-05-08
new sample

sheepdreamofandroids:
It took a while, but here's the sample of batching loads for #309

### DIFF
--- a/examples/coalescing-bulkloader/pom.xml
+++ b/examples/coalescing-bulkloader/pom.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.github.benmanes.caffeine.examples</groupId>
+  <artifactId>coalescing-bulkloader</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <java.version>1.8</java.version>
+    <caffeine.version>2.7.0</caffeine.version>
+    <junit.version>4.12</junit.version>
+    <awaitility.version>3.1.6</awaitility.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+      <version>${caffeine.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>${junit.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>${awaitility.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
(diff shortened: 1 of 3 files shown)


ben-manes:
Thanks @sheepdreamofandroids!

I'm curious about the `load(Map)` method. From a user's perspective, I imagine it might be cleaner to return the futures instead of consume them. There could be all sorts of bugs, such as that method throwing an exception. I imagine in practice you have more edge case handling and tried to be minimal here for the clearest demonstration? My concern is that someone might just copy it as-is and get bitten.

sheepdreamofandroids:
Well, I wrote this from scratch since it turned out that there are just too many business logic concerns in the current code. I put this PR out here to get exactly this kind of feedback. 

load() has this signature because the underlying bulk call usually associates keys with values in one of two ways: either it returns each value together with the key, or it returns the values in the same order as the keys.

I tried to make both ways possible in the same call without forcing the user to create a map or reorder values.

Thinking about it, I could probably allow different lambda's to be passed in the constructor to allow for both ways.

Maaartinus:
> either it returns each value together with the key, or it returns the values in the same order as the keys.

Another pretty common case may be returning the values unsorted, but knowing how to get the key from the value. This is the case with simple `SELECT ... FROM ... WHERE id in (....)`.

sheepdreamofandroids:
> Another pretty common case may be returning the values unsorted, but knowing how to get the key from the value. This is the case with simple `SELECT ... FROM ... WHERE id in (....)`.

True. And now that I can have different types of wrapped loaders, nothing stops me from supporting that explicitly ;-)

sheepdreamofandroids:
> Another pretty common case may be returning the values unsorted, but knowing how to get the key from the value. This is the case with simple `SELECT ... FROM ... WHERE id in (....)`.

@Maaartinus Is this what you had in mind?

ben-manes:
Sorry that I haven't had the time to merge this. It's on my backlog, so maybe I'll have the energy next weekend to work on these things.

sheepdreamofandroids:
No worries!

Life just won't adapt to GitHub...

ben-manes:
Sorry for the long delay. I had been considering to refactor things a bit in the handoff, but that was leaving this linger. So I tried to merge it and run, but unfortunately the tests failed.

```
Running com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest
whenEnoughKeysAreRequestedTheLoadWillHappenImmediately[0](com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest) failed, 2 attempts left.
maxDelayIsNotMissedTooMuch[2](com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest) failed, 2 attempts left.
maxDelayIsNotMissedTooMuch[2](com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest) failed, 1 attempts left.
maxDelayIsNotMissedTooMuch[2](com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest) failed, 0 attempts left.
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.706 sec <<< FAILURE!
maxDelayIsNotMissedTooMuch[2](com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest)  Time elapsed: 0.215 sec  <<< FAILURE!
java.lang.AssertionError:
Expected: is <1>
     but: was <0>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at com.github.benmanes.caffeine.examples.coalescing.bulkloader.CoalescingBulkloaderTest.maxDelayIsNotMissedTooMuch(CoalescingBulkloaderTest.java:98)
```

ben-manes:
Sorry, this was closed by accident and I can't reopen..

PR #317, opened 2019-05-10
Blocked Bloom filter (faster than regular Bloom; fpp fixed at around 1%)

thomasmueller:
This filter should be fast than a regular Bloom filter. I think you use a fpp of 0.1, that is 10%. That's unusually high for a Bloom filter... I'm afraid I don't understand this part. My BlockedBloom uses 0.01 (1%). Sure, I can make a 10% fpp filter, but I'm I would like to understand the reason to use 10%.

I'm afraid I don't know how to test it, even running the existing ones doesn't work for me:

    ./gradlew simulator:run

> Task :simulator:run
[ERROR] [05/10/2019 16:05:20.380] [Main-akka.actor.default-dispatcher-2] [akka.actor.ActorSystemImpl(Main)] 
java.lang.NumberFormatException: For input string: "version https://git-lfs.github.com/spec/v1"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Long.parseLong(Long.java:589)
        at java.lang.Long.parseLong(Long.java:631)
        at java.util.stream.ReferencePipeline$5$1.accept(ReferencePipeline.java:227)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1812)
        at java.util.stream.StreamSpliterators$LongWrappingSpliterator.lambda$initPartialTraversalState$0(StreamSpliterators.java:405)
        at java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.fillBuffer(StreamSpliterators.java:206)
        at java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.doAdvance(StreamSpliterators.java:161)
        at java.util.stream.StreamSpliterators$LongWrappingSpliterator.tryAdvance(StreamSpliterators.java:416)
        at java.util.stream.StreamSpliterators$LongWrappingSpliterator.tryAdvance(StreamSpliterators.java:379)
        at java.util.stream.Streams$ConcatSpliterator$OfPrimitive.tryAdvance(Streams.java:799)
        at java.util.stream.Streams$ConcatSpliterator$OfLong.tryAdvance(Streams.java:823)
        at java.util.Spliterators$3Adapter.hasNext(Spliterators.java:771)
        at com.github.benmanes.caffeine.cache.simulator.Simulator.broadcast(Simulator.java:105)

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/membership/FilterType.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/membership/FilterType.java
@@ -15,6 +15,7 @@
  */
 package com.github.benmanes.caffeine.cache.simulator.membership;
 
+import com.github.benmanes.caffeine.cache.simulator.membership.bloom.BlockedBloom;
 import com.github.benmanes.caffeine.cache.simulator.membership.bloom.BloomFilter;
 import com.github.benmanes.caffeine.cache.simulator.membership.bloom.GuavaBloomFilter;
 
@@ -39,7 +40,16 @@ public enum FilterType {
     @Override public String toString() {
       return "Guava";
     }
-  };
+  },
+  BLOCKED_BLOOM {
+      @Override public Membership create(long expectedInsertions, double fpp) {
+        return new BlockedBloom(expectedInsertions, fpp);
+      }
+      @Override public String toString() {
+        return "BlockedBloom";
+      }
+    };
+  
 
   /**
    * Returns a new membership filter.
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/membership/bloom/BlockedBloom.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/membership/bloom/BlockedBloom.java
@@ -1,0 +1,57 @@
+package com.github.benmanes.caffeine.cache.simulator.membership.bloom;
+
+import java.util.Arrays;
+
+import com.github.benmanes.caffeine.cache.simulator.membership.Membership;
+
+public class BlockedBloom implements Membership {
+
+    private final int buckets;
+    private final long seed;
+    private final long[] data;
+
+    public BlockedBloom(long expectedInsertions, double fpp) {
+        // the fpp parameter is ignored; fpp is about 1%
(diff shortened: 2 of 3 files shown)


thomasmueller:
By the way, the BlockedBloom filter basically a copy of this one: https://github.com/FastFilter/fastfilter_java/blob/master/src/main/java/org/fastfilter/bloom/BlockedBloomV2.java

ben-manes:
Oh, that fpp was a typo. :)

I moved the trace files to git lfs which means it’s just stubs if you don’t have that installed to pull it down. Otherwise you would need to set a new file path to a trace you downloaded separately.

We could pull in your repo’s using jitpack.io instead, would that be better?

Thanks for this. 😄

ben-manes:
Thanks!

Sorry for the confusion wrt running the simulator. If you install git lfs
<https://git-lfs.github.com/> then it should pull them down automatically.
That's usually done on a clone, but a `git lfs pull` or similar might
resolve it correctly after the fact.

If we use jitpack.io then it can build your repository and provides a
dependency for Gradle. That way if you made improvements or other
implementations, we can just change the commit hash to upgrade. The only
limitation would need to be implement clear() like in the Guava adapter.

What do you think?

ben-manes:
@thomasmueller I have this refactored to use jitpack and support your other bloom filter types generically. A bunch of other random changes locally so I need to break things apart when I clean up the commit. I'll close this when I've merged, thanks!

ben-manes:
If you checkout with git-lfs or download one of the external traces (ARC, SNIA, UMass), you can run the simulator with FastFilter. I refactored the configuration to be more flexible and reuse the adapter so that we can run any of your filters that support adds. See the [chunk](https://github.com/ben-manes/caffeine/commit/1b5122334eeeae6d2a662c32e356eaa7784aa28a#diff-df0ee9ac0607232cf0973446a7c7072d) if interested.

I added you as a co-author for the commit. 😺

PR #333, opened 2019-08-01
Change method name 'counter' to 'generateNumbers'

pdhung3012:
The class Synthetic is used to generate Synthetic cache event.  This method named 'counter' is to generate sequence of unique integers. Thus, the method name 'generateNumbers' is more intuitive than 'counter'.


ben-manes:
Thanks. 😄 

I used `Counter` since that is the [YCSB](https://github.com/brianfrankcooper/YCSB) distribution generator's name and, technically, all of them generate numbers! So in that sense, I don't think the new name is an improvement.

PR #335, opened 2019-08-02
Add verification of plugins and dependencies via com.github.vlsi.checksum-dependency

vlsi:
See https://github.com/ben-manes/caffeine/issues/301#issuecomment-467731577

This enables to verify checksums before they got used (e.g. before plugins are activated), so it would prevent `stealing a coordinate` issues.

PR #342, opened 2019-09-03
Use PGP for dependency verification

vlsi:
See https://github.com/vlsi/vlsi-release-plugins/tree/master/plugins/checksum-dependency-plugin

This is a follow-up to https://github.com/ben-manes/caffeine/pull/335

TL;DR: updated xml is generated to `build/checksum/checksum.xml`.
`-PchecksumUpdate` makes the plugin to update `$rootDir/checksum.xml` directly.

By default it tries to use group PGP keys, however if no key is present, then it would add section like (that would mean the plugin would stop trying PGP for those dependencies)

```xml
<dependency group='org.ow2.asm' module='asm-util' version='7.0'>
  <sha512>6B9630AD4C5F6ABC2724F9D85A7C0E81E524F8B0BF25273DD5578EDC20EFC7AB400095CD48E904D2536071181A8E9F39635FF3FEE0986BD078016D2DFF8241B1</sha512>
</dependency>
```

It turns out `com.beust:jcommander:1.78` is signed by `ac5ec74981f9cda6`, however public part of that key is not available, so the signature can't be verified. That is why `<ignored-key id='ac5ec74981f9cda6' />` is added.

PS. I've generated `checksum.xml` file with `CI=Y gw allDependencies build -x test -x compileJava -x compileTestJava -x compileJmhJava -x javadoc -PchecksumUpdate -PchecksumFailOn=never`


vlsi:
@ben-manes it would be nice if you could review this change.

However, it looks like [PGP keyserver failover](https://github.com/vlsi/vlsi-release-plugins/issues/11) is crucial.
Sample failure: https://travis-ci.org/ben-manes/caffeine/jobs/580123737#L2176

ben-manes:
Would switching to okhttp resolve this?

vlsi:
>Would switching to okhttp resolve this?

okhttp would solve this at a cost of adding extra dependencies.

A mere adding of `classpath(com.squareup.okhttp3:okhttp:4.1.0)` produces

```
Buildscript classpath has non-whitelisted files:
    SHA-512(annotations-13.0.jar) = 5622D0... (.../org.jetbrains/annotations/...)
    SHA-512(kotlin-stdlib-1.3.41.jar) = B0F79A... (.../org.jetbrains.kotlin/kotlin-stdlib/...)
    SHA-512(kotlin-stdlib-common-1.3.41.jar) = 2EC60C (.../org.jetbrains.kotlin/kotlin-stdlib-common/...)
    SHA-512(okhttp-4.1.0.jar) = 17DAAF... (.../com.squareup.okhttp3/...)
    SHA-512(okio-2.2.2.jar) = 93E7A4... (.../com.squareup.okio/...)
```

Current Gradle should include `kotlin-stdlib`, so `classpath("com.squareup.okhttp3:okhttp:4.1.0") { exclude("org.jetbrains.kotlin", "kotlin-stdlib")  }` would probably work.

`okhttp`/`okio` might be a fair set of extra dependencies.

ben-manes:
There are many others, like Apache's httpcomponents, which might be slimmer.

vlsi:
>Would switching to okhttp resolve this?

There's a side feature: support of multiple cross DNS failover (== ability to use multiple different DNS names for resolution)

This seems to be relevant:
https://github.com/phar-io/phive/commit/92f9642233d48e0d4b6cf5e2701b1e13091d0b35#diff-5e5b2bcab99b42b28d693831aa29d429
and
https://github.com/phar-io/phive/commit/92f9642233d48e0d4b6cf5e2701b1e13091d0b35#diff-5e5b2bcab99b42b28d693831aa29d429

vlsi:
Just in case, `HttpURLConnection` forbids to use `Host: ...` header for an unknown security reason.

In other words, `java.net.URL#openConnection()` can't really be used for DNS load-balancing.

vlsi:
Ok, so I've switched to okhttp, and it turns out key resolution still takes time, so it is probably worth resolving keys in parallel:

https://travis-ci.org/ben-manes/caffeine/jobs/581576548#L270

    PGP key resolution time: 23975ms, resolution requests: 146, download time: 0ms, keys downloaded: 86
    PGP verification time: 1240ms, files processed: 146, processed: 107MiB, skipped: 77MiB


https://travis-ci.org/ben-manes/caffeine/jobs/581576548#L321

    PGP key resolution time: 11150ms, resolution requests: 218, download time: 0ms, keys downloaded: 26
    PGP verification time: 1290ms, files processed: 218, processed: 145MiB, skipped: 641MiB

ben-manes:
nice. Let me know when it's no longer WIP.

vlsi:
I find 1.24.0 is good enough. Parallel PGP resolver would take some time to implement, so I would suggest to merge current version and collect some statistics on how it works.

I've filed PRs to multiple projects, and it so far it receives PGP keys just fine.

There's a minor glitch in the logging: `PGP... download time: 0ms` is always printed as 0 (which is fixed in https://github.com/vlsi/vlsi-release-plugins/commit/3deb95aede2ee1da962875a2bbfe605f47bf1a7f )
However it looks a bit odd to release a version with logging fix only.

vlsi:
I've updated checksum-dependency-plugin to 1.26.0, and I'm quite confident it is good to go.
I've improved logging, so in case of failure it prints the location of the updated `checksum.xml` file, so I hope that would reduce RTFM/day ratio.

ben-manes:
Thanks, I'll play with this over the weekend.

ben-manes:
Sorry, I caught the flu and got behind on this. I played with it and I appreciate that you added `checksumIgnore` and `checksumUpdate`.

We could now remove the conditional apply in the `settings.gradle`, add `checksumIgnore=true` in `gradle.properties`, and the environmental flag `ORG_GRADLE_PROJECT_checksumIgnore=false` to the CI build. I set a local override in `~/.gradle/gradle.properties` to enable the plugin.

This way anyone who clones this project can hack it without being impacted if they add their own dependencies, etc. The CI will validate so that we know that our builds are trustworthy and my local override ensures that it is run when upgrading dependencies.

This plugin interacts poorly with the `gradle-versions-plugin` by rejecting its ability to resolve dynamic dependencies. For your users you might consider detecting if `dependencyUpdates` is on the task graph to disable automatically. Otherwise one needs to run `gradle dU -PchecksumIgnore=true` to get the report.

PR #349, opened 2019-09-23
Nit: Update the Javadocs of `refreshAfterWrite` to reflect the actual beha…

jyterencekim:
…vior for IllegalArgumentException

## Summary
- The current implementation and the current [test code](https://github.com/ben-manes/caffeine/blob/19f0911a78fa8096a4f610a24b756741dd4ed984/caffeine/src/test/java/com/github/benmanes/caffeine/cache/CaffeineTest.java#L513-L516) imply that `refreshAfterWrite` throws an `IllegalArgumentException` **also** when `duration` is zero. This pull request clarifies this fact in the Javadoc.


ben-manes:
Thanks!

PR #353, opened 2019-09-29
Bump checksum-dependency-plugin to 1.30.0

vlsi:
It disables the plugin in case `dependencyUpdates` is on the task execution graph (the set of tasks is configurable via `checksumIgnoreOnTask` property)


ben-manes:
perfect!

PR #361, opened 2019-11-05
chore: upgrade dependencies.

croudet:


croudet:
```
Execution failed for task ':caffeine:uploadArchives'.

> Could not publish configuration 'archives'

   > Could not deploy to remote repository | Failed to deploy artifacts: Could not transfer artifact com.github.ben-manes.caffeine:caffeine:jar:2.8.1-20191105.151335-21 from/to remote (https://oss.sonatype.org/content/repositories/snapshots/): Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/com/github/ben-manes/caffeine/caffeine/2.8.1-SNAPSHOT/caffeine-2.8.1-20191105.151335-21.jar. Return code is: 401, ReasonPhrase: Unauthorized.
```
I don't know how to fix that.

gradle/dependencies.gradle
```diff
@@ -49,51 +49,51 @@ ext {
...
     easymock: '4.0.2',
-    hamcrest: '2.1',
+    hamcrest: '2.2',
     jcacheTck: '1.1.1',
     jctools: '2.1.2',
     junit: '4.12',
     mockito: '3.1.0',
     paxExam: '4.13.1',
-    testng: '6.14.3',
+    testng: '7.0.0',
```
ben-manes: I had memory issues with 7.0, so I am surprised it's working for you. I'd lean towards staying at 6.x for now to be safe.

See https://github.com/cbeust/testng/issues/2096 where I spent time trying to reduce the memory churn to compensate for TestNG's added bloat. They don't spool the test results to disk and instead hold it in memory, which doesn't scale for our test execution count (millions).

ben-manes:
You'll probably need to [update the checksums](https://github.com/vlsi/vlsi-release-plugins/blob/master/plugins/checksum-dependency-plugin/README.md), we is strict on CI for the master branch. The validation is disabled for local to not be bothersome.

You can enable it locally by adding `-PchecksumIgnore=false` locally or in `~/.gradle/gradle.properties`. If you run with `-PchecksumUpdate` then it will rebuild the manifest with the upgrades.

ben-manes:
oh, I guess that bash should be an AND condition? If CI and not PR. wdyt?

ben-manes:
PMD can also be upgraded. The `gradlew dependencyUpdates` report doesn't include it since it's a built into Gradle. The latest version passes `./travis.sh analysis` with a minor warning of a renamed rule,

>Use Rule name category/java/errorprone.xml/InvalidLogMessageFormat instead of the deprecated Rule name category/java/errorprone.xml/InvalidSlf4jMessageFormat. PMD 7.0.0 will remove support for this deprecated Rule name usage.

That can be could by excluding the old rule in `ruleSets.xml`, or ignored and wait for v7 to remove it as its just console noise.

vlsi:
@ben-manes , AFAIK Travis does not run checksum-dependency (see https://github.com/ben-manes/caffeine/pull/362 )

For instance:

https://travis-ci.org/ben-manes/caffeine/jobs/607844055#L228
```
checksum-dependency-plugin is disabled since checksumIgnore property is present
```

It looks like there are mismatches in the current `checksum.xml` configuration: https://travis-ci.org/vlsi/vlsi-release-plugins/jobs/603217203#L453 (I've a Travis job that verifies that updated c-d-p does not break Caffeine's master)

ben-manes:
Odd. I have it set as an environmental variable. I thought it would be picked up then?

Screen Shot 2019-11-05 at 1 26 08 PM

vlsi:
secure environment variables are not injected to pull request jobs.
Then Caffeine's Travis jobs seem to skip `checkstyle` (see https://travis-ci.org/ben-manes/caffeine/jobs/595453279#L270 ), thus it does not trigger checksum-dependency-verification failure even though Caffeine's master is "verified" in Travis.

^^ I hope the above is understandable.

ben-manes:
oh, I've never enable checkstyles as enforcing strict formatting rules was a little too prescriptive. It's configured if we decided that was useful, though.

That makes to not inject them into PRs and I simply hadn't thought about it

croudet:
For PMD, excludind the deprecated rule does not clear the warning, there is an opened issue about that:
https://github.com/pmd/pmd/issues/2096

ben-manes:
thanks for checking! Let's leave the warning and take the upgrade.

ben-manes:
beautiful, it passed! Thanks for the PR and working through the issues. 😄

PR #363, opened 2019-11-09
Bump checksum-dependency-plugin -> 1.44, fix allDependencies

vlsi:
v1.31.0, v1.33.0, v1.35.0 reduce the verbosity when downloading keys.
v1.31.0 adds `-PchecksumUpdateAll` property (==update `checksum.xml`, avoid failures)

I use `caffeine / allDependencies` for regression tests of `checksum-dependency-plugin`, so I would like the task to be working: https://travis-ci.org/vlsi/vlsi-release-plugins/jobs/609652585#L444-L444


ben-manes:
Thanks! I was doing a few other improvements so rebased my work on yours.

PR #365, opened 2019-11-20
Make BoundedBuffer a public class so that it can be used in different…

marianogonzalez:
… contexts

This is a simple request to make the BoundedBuffer class public. I found that class to be quite helpful in scenarios other than a LoadingCache and it'd be great to have it as part of Caffeine's offering


ben-manes:
I'm glad that you found it useful. However, I think that over time we'll further restrict any non-cache related utilities from being public. You're welcome to copy the code out and use it in your projects.

[JCTools](https://github.com/JCTools/JCTools) does a really nice job offering data structures like this and we've borrowed from it ourselves. You might check with @nitsanw to see if he'd like to provide something fulfilling your needs more directly. That might be to pull this out or design something from scratch that he things is more suitable.

PR #374, opened 2019-12-09
Extend Caffeine Simulator to handle different requests types

ohadeytan:
We aim to extend Caffeine Simulator to handle different requests types, such as size-aware and latency-aware requests. 
To do so, we first create an AccessEvent class that will handle all the requests during parsing and processing. For now, it handles only regular key accesses. 
This is still WIP, but I opened the pull-request so you can review and comment along the way.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/Simulator.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/Simulator.java
@@ -21,6 +21,7 @@ import static com.github.benmanes.caffeine.cache.simulator.Simulator.Message.STA
 import static java.util.stream.Collectors.toList;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.PrimitiveIterator;
@@ -104,12 +105,12 @@ public final class Simulator extends AbstractActor {
   /** Broadcast the trace events to all of the policy actors. */
   private void broadcast() {
     try (Stream<AccessEvent> events = eventStream()) {
-      LongArrayList batch = new LongArrayList(batchSize);
+      ArrayList<AccessEvent> batch = new ArrayList<AccessEvent>(batchSize);
       for (Iterator<AccessEvent> i = events.iterator(); i.hasNext();) {
-        batch.add(i.next().getKey());
+        batch.add(i.next());
         if (batch.size() == batchSize) {
           router.route(batch, self());
-          batch = new LongArrayList(batchSize);
+          batch = new ArrayList<AccessEvent>(batchSize);
         }
       }
       router.route(batch, self());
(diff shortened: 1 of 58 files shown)


ben-manes:
One trick to minimize impacts can be to use adapters via base classes / default interfaces. For example,

```java
public interface TraceReader {
  Set<Policy.Characteristics> characteristics();
  Stream<AccessEvent> events() throws IOException;

  interface KeyOnlyTraceReader extends TraceReader {
    default Set<Policy.Characteristics> characteristics() {
      return ImmutableSet.of();
    }

    LongStream keys() throws IOException;

    @Override
    default Stream<AccessEvent> events() throws IOException {
      return keys().mapToObj(AccessEvent::new);
    }
  }
}
```

Then `ArcTraceReader` would include `implements KeyOnlyTraceReader` and rename `events()` to `keys()`. That's about the same amount of work, but lets you get rid of the boilerplate of the empty characteristics method in each case. The same could be done for the `Policy` to make migrations more trivial.

I would also create `AccessEvent` by static factory methods, like `forKey(long key)` and `forKeyAndWeight(long key, int weight)`. This would let us implement specialized versions with only the fields necessary, rather than a single type where they are defaulted as unused. Since immutable, drop the get prefix as unnecessary.

ben-manes:
I had some of this started yesterday. If you don't mind, I could wrap it up tomorrow for you. I wanted to bring in one of the sized traces for verifying the changes with.

ohadeytan:
Great. I don't think I will touch it in the meantime.

ben-manes:
```
format = adapt-size
paths = [ request.trace.xz ]
maximum-size = 51200
```

This filters out to only the policies that support the weight characteristic. This was enabled for the vendors that offer this feature.

```
╔═══════════════════════╤══════════╤═══════╤════════╤══════════╤═══════════╤════════════╤═══════╤══════════╗
║ Policy                │ Hit rate │ Hits  │ Misses │ Requests │ Evictions │ Admit rate │ Steps │ Time     ║
╠═══════════════════════╪══════════╪═══════╪════════╪══════════╪═══════════╪════════════╪═══════╪══════════╣
║ product.Cache2k       │ 4.81 %   │ 3,223 │ 63,764 │ 66,987   │ 0         │ 100.00 %   │ ?     │ 468.0 ms ║
╟───────────────────────┼──────────┼───────┼────────┼──────────┼───────────┼────────────┼───────┼──────────╢
║ product.Caffeine      │ 11.88 %  │ 7,955 │ 59,032 │ 66,987   │ 0         │ 100.00 %   │ ?     │ 393.1 ms ║
╟───────────────────────┼──────────┼───────┼────────┼──────────┼───────────┼────────────┼───────┼──────────╢
║ product.ElasticSearch │ 7.97 %   │ 5,340 │ 61,647 │ 66,987   │ 61,643    │ 100.00 %   │ ?     │ 181.7 ms ║
╟───────────────────────┼──────────┼───────┼────────┼──────────┼───────────┼────────────┼───────┼──────────╢
║ product.Guava         │ 9.77 %   │ 6,546 │ 60,441 │ 66,987   │ 60,423    │ 100.00 %   │ ?     │ 260.3 ms ║
╚═══════════════════════╧══════════╧═══════╧════════╧══════════╧═══════════╧════════════╧═══════╧══════════╝
```

PR #375, opened 2019-12-09
Further extension for simulator - penalties (latency aware)

himelbrand:
I made the following additions/changes:

- Added support for PolicyStats to use AccessEvent so it will be able to use the metadata to gather additional stats. (added latency stats)

- Extended the Reporter to show extra information when relevant, i.e. show latency stats when using relevant traces. (uses the characteristics defined in the config file under report section)

- Added another attribute in the report section in config file, for time-unit, so the report will show the relevant latency.

- Edited the Registry filter to allow running policies that do not use certain characteristics, as long as their set of characteristics is a subset of the trace reader set of characteristics.

- Added PENALTIES characteristic as enum in Policy

- Added parser with PENALTIES characteristic that reads extended traces of address format (the same format with added hit and miss penalties) I also added an example trace.

- Extended AccessEvent


ben-manes:
If you are using a text editor, like Visual Code, then you might prefer Eclipse or IntelliJ. I use the former. We use Google's Java code style and the formatter is provided in their [repository](https://github.com/google/styleguide). It looks like you reformatted and wrote using a different style. Also please remove unnecessary empty lines, etc.

There appears to be a lot of unnecessary changes, e.g. `ClockProPolicy`. Why does it need to use the `AccessEvent` rather than remain based on the `long` key? Only those policies using an `admittor` seem like they need to change to be aware of the access event. And even then, only very slightly when calling `record(key)` to become `record(event)`.

Is there a need to have the policy stats and report not include the penalties by default, thereby not passing in the characteristics to do extra formatting? I think it's okay to just have additional columns in the output. The changes on `PolicyStats` to record the penalties is understandable, but also impactful. Ideally we'd be able to do something less painful, as it is more invasive than I was hoping.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Registry.java
```diff
@@ -100,7 +100,7 @@ private Registry() {}
   public static Set<Policy> policies(BasicSettings settings, Set<Characteristic> characteristics) {
     return settings.policies().stream()
         .flatMap(name -> policy(settings, name).stream())
-        .filter(policy -> policy.characteristics().containsAll(characteristics))
+        .filter(policy -> characteristics.containsAll(policy.characteristics()))
```
ben-manes: Why is this inverted?

The prior logic said that the policy must implement the minimum characteristics, as defined by the trace. This allowed the policy to be weighted and the trace not be, so the default weight of `1` would be supplied and compatible.

The new logic says that the policy must implement a subset of the trace's characteristics.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/AccessEvent.java
```diff
@@ -27,69 +27,122 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        } else if (!(o instanceof AccessEvent)) {
+            return false;
+        }
+        AccessEvent event = (AccessEvent) o;
+        return Objects.equals(key(), event.key())
```
ben-manes: All of the new fields should be added to the equals/hashCode/toString

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/address/penalties/AddressPenaltiesTraceReader.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public Set<Policy.Characteristic> characteristics() {
+    return Sets.immutableEnumSet(PENALTIES);
+  }
+
+  @Override
+  public Stream<AccessEvent> events() throws IOException {
+    return lines()
```
ben-manes: Why split and join instead of extracting the three array elements?

```java
return lines()
  .map(line -> line.split(" ", 5))
  .map(split -> AccessEvent.forKeyAndPenalties(Long.parseLong(split[1], 16),
    Integer.parseInt(split[3]), Integer.parseInt(split[4])));
```

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Policy.java
```diff
@@ -48,10 +49,6 @@ default void finished() {}
       return ImmutableSet.of();
     }
 
-    @Override default void record(AccessEvent event) {
```
ben-manes: This defeats the purpose of having this adapter. Either the change is wrong or the adapter has to be removed. It was introduced to make it less invasive for existing policies who don't care about the full event.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Registry.java
himelbrand: I see what you're saying, but I had my extension of a trace with penalties, where I would like for policies that do not use this characteristic to still run, so it will be possible to see how they compare by the metric of QoS/latency.
a possibility to handle this issue can be to add an attribute to the config file that will decide if it should be: the prior logic, the new logic, and maybe another option of equality of the sets.

BTW, a check if the number of policies that remain after filtering in the registry is > 0 is needed, else the simulator won't stop. (I'll add this now)

let me know what you think about my suggestion regarding the config attribute to handle the filtering logic, so it would be able to fit multiple simulation requirements.

ben-manes: It sounds like this isn't an optional characteristic then, because you want all policies to report latency information. This means that, like `key`, it is fundamental to a policy to handle. What is the benefit of having the characteristic then? It merely says whether the algorithm uses it in its decisions, which is not very meaningful (as they don't report whether they use recency or frequency, for instance). If that's the case, then I don't think you should add a characteristic.

This would explain why the changes were more invasive than I expected. It would be nice to not have it impact every one written, though perhaps there is no workaround. The latencies could change over the lifetime workload (e.g. network getting slower when congested), which could cause some surprising skew based on when the miss occurred.

ben-manes: One approach might be to add the penalty in `PolicyActor`, like we do for the execution time. If the miss count increased, add the miss penalty. If the hit count increased, that one. If neither, ignore (e.g. opt must wait for its replay). That's a little less invasive as writing a policy won't need to account for the QoS metrics.

himelbrand: Right now, there are no implemented policies that use this information for decisions, but I intend on adding such a policy, perhaps even more than one.
And, when using it I might want to compare it's stats to other policies that might not use this information in the decision-making process.

regarding the changes in the miss penalties over time, I am aware of that, and the algorithm I'm working on tries to take that into consideration, in addition, we are now working on traces from a real system. 

so if a data from the trace is needed for some of the simulations, I'm not sure how would moving it to the PolicyActor will help.

ben-manes: It would move the accounting in the stats. Currently every policy is modified to record the stats with your metrics. Instead you could record the penalty time outside of those policies and not affect them. If you need it in a policy, it can use it as it’s available in the event as desired.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/address/penalties/AddressPenaltiesTraceReader.java
himelbrand: I fixed this in the latest push

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/AccessEvent.java
himelbrand: Fixed in the latest push

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Registry.java
himelbrand: okay, I'll try and change it now.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Policy.java
himelbrand: I'll change it back, and it should work fine with the change to PolicyActor

himelbrand: changed back in the last commit

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/Registry.java
himelbrand: Okay, I changed the penalties to work via the PolicyActor as you suggested, but still, I do think that the logic of the filtering should be the way I did it since it keeps only the policies that will work properly with the type of event that the trace reader provides and allows comparing results against policies that do not factor in all of the provided meta-data to make a decision.

If you think this filtering is not appropriate for all use cases, I think that the filtering logic should be controlled in some way, my current idea is still to add some attribute to the configuration file to control the filtering logic.

currently, I don't really see a reason we would want to filter out policies that will work properly when using some trace reader.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/AccessEvent.java
himelbrand: in order to pass the travis.sh analysis, I needed to override equals and hashCode inside the PenaltiesAccessEvent class.
when I just added the fields to the methods that are defined in AccessEvent the analysis failed due to:
EQ_DOESNT_OVERRIDE_EQUALS: Class doesn't override equals in superclass

if you have another idea how to overcome this warning I could make the changes, but right now that's the only way I was able to make this warning go away.

himelbrand:
> If you are using a text editor, like Visual Code, then you might prefer Eclipse or IntelliJ. I use the former. We use Google's Java code style and the formatter is provided in their [repository](https://github.com/google/styleguide). It looks like you reformatted and wrote using a different style. Also please remove unnecessary empty lines, etc.
> 
> There appears to be a lot of unnecessary changes, e.g. `ClockProPolicy`. Why does it need to use the `AccessEvent` rather than remain based on the `long` key? Only those policies using an `admittor` seem like they need to change to be aware of the access event. And even then, only very slightly when calling `record(key)` to become `record(event)`.
> 
> Is there a need to have the policy stats and report not include the penalties by default, thereby not passing in the characteristics to do extra formatting? I think it's okay to just have additional columns in the output. The changes on `PolicyStats` to record the penalties is understandable, but also impactful. Ideally we'd be able to do something less painful, as it is more invasive than I was hoping.

I'm using Intellij, I didn't notice the part about the formatting, I did it now, using the google-java-formatter.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/AccessEvent.java
ben-manes: Yes, please add it to the exclusions instead, which for Spotbugs is in `config/excludes.xml`:

https://github.com/ben-manes/caffeine/blob/aa0cdfad2e94e4a39432f7d27f43ff0a3e25755a/config/spotbugs/exclude.xml#L140-L143

himelbrand: No problem, doing it now

himelbrand:
Changed penalties type from int to double, after seeing a few real traces with penalties that were floating-point numbers.
This way both types of traces will be handled.

ben-manes:
Thanks for putting all of this together and including a trace.

There are a lot of changes (100 files), so going back-and-forth would probably be frustrating for both of us. Since this is impacting a lot of shared code, I'm more guarded than if you were writing a new policy where anything in your new files would be treated more akin to a free-for-all. In those cases I'd respect the research aspects where, as long as it was stable and didn't harm anything, I'd have only cosmetic touch-ups to follow existing conventions. In this case it has wide breadth on shared code that I use and maintain.

A better path forward might be if I pick it up and rework it to fit how I think the simulator should evolve. Then hopefully it fits both our needs and we can absorb this and future contributions. If that sounds okay, then I'd try to do that over the weekend where I should have some free cycles. It may be that after getting my hands dirty, I'm more inclined to how you tackled things than I was originally.

Worst case, if we didn't get this upstream you won't be blocked on your goals by working in your fork. So there is a benefit that as the long-term owner I can have a high bar on what is accepted, but as a contributor you can ignore me and move forward independently. I would like to see this get in, hence my willingness to jump in to make sure it's something I feel comfortable having to maintain.

PR #378, opened 2019-12-16
Make linked policies simulation weight-aware

ohadeytan:
I changed the linked policies (LRU, MRU, FIFO, Clock) to be able to work with weighted traces.
I validate that they work the same as before for unweighed traces, and the new implementation has identical results to [webcachesim](https://github.com/dasebe/webcachesim) simulator (for LRU and FIFO).

Notice that if the admission policy is not Always, the behavior is to evict victims one by one as long as the candidate wins and admit when there is enough space (similar to [Ristretto](https://github.com/dgraph-io/ristretto) approach and not to Caffeine current approach).

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/PolicyStats.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/PolicyStats.java
@@ -83,6 +83,7 @@ public final class PolicyStats {
 
   public void recordWeightedHit(int weight) {
     hitsWeight += weight;
+    recordHit();
   }  
   
   public long hitsWeight() {
@@ -111,6 +112,7 @@ public final class PolicyStats {
 
   public void recordWeightedMiss(int weight) {
     missesWeight += weight;
+    recordMiss();
   }  
   
   public long missesWeight() {
@@ -184,7 +186,6 @@ public final class PolicyStats {
     long requestsWeight = requestsWeight();
     return (requestsWeight == 0) ? 1.0 : (double) missesWeight / requestsWeight;
   }
-
 
   public double admissionRate() {
     long candidateCount = admittedCount + rejectedCount;
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/PolicyStats.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/PolicyStats.java
@@ -31,6 +31,8 @@ public final class PolicyStats {
   private String name;
   private long hitCount;
   private long missCount;
+  private long hitsWeight;
+  private long missesWeight;
   private double hitPenalty;
   private double missPenalty;
   private long evictionCount;
@@ -79,6 +81,14 @@ public final class PolicyStats {
     hitCount += hits;
   }
 
+  public void recordWeightedHit(int weight) {
+    hitsWeight += weight;
+  }  
+  
+  public long hitsWeight() {
+    return hitsWeight;
+  }
+  
   public void recordHitPenalty(double penalty) {
     hitPenalty += penalty;
   }
@@ -99,6 +109,14 @@ public final class PolicyStats {
     missCount += misses;
   }
 
+  public void recordWeightedMiss(int weight) {
+    missesWeight += weight;
+  }  
+  
+  public long missesWeight() {
+    return missesWeight;
+  }
+  
   public void recordMissPenalty(double penalty) {
     missPenalty += penalty;
   }
@@ -121,6 +139,10 @@ public final class PolicyStats {
 
   public long requestCount() {
     return hitCount + missCount;
+  }
+
+  public long requestsWeight() {
+    return hitsWeight + missesWeight;
   }
 
   public long admissionCount() {
@@ -148,10 +170,21 @@ public final class PolicyStats {
     return (requestCount == 0) ? 1.0 : (double) hitCount / requestCount;
   }
 
+  public double weightedHitRate() {
+    long requestsWeight = requestsWeight();
+    return (requestsWeight == 0) ? 1.0 : (double) hitsWeight / requestsWeight;
+  }
+
   public double missRate() {
     long requestCount = requestCount();
     return (requestCount == 0) ? 0.0 : (double) missCount / requestCount;
   }
+
+  public double weightedMissRate() {
+    long requestsWeight = requestsWeight();
+    return (requestsWeight == 0) ? 1.0 : (double) missesWeight / requestsWeight;
+  }
+
 
   public double admissionRate() {
     long candidateCount = admittedCount + rejectedCount;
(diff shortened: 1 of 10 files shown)


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
```diff
@@ -66,19 +73,31 @@ public LinkedPolicy(Admission admission, EvictionPolicy policy, Config config) {
...
   public PolicyStats stats() {
     return policyStats;
   }
-
+  
   @Override
-  public void record(long key) {
+  public void record(AccessEvent event) {
+    final int weight = event.weight();
+    final long key = hashCode(event.key(), weight);
```
ben-manes: Why does the key carry the weight?

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
```diff
@@ -240,8 +264,14 @@ public void moveToTail() {
...
       return MoreObjects.toStringHelper(this)
           .add("key", key)
+          .add("weight", weight)
           .add("marked", marked)
           .toString();
     }
   }
+  
+  /** Cantor pairing function. */
+  private long hashCode(long key, int weight) {
```
ben-manes: fyi, `Objects.hashCode(...)` is a convenience for these cases

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
```diff
@@ -89,22 +108,25 @@ public void record(long key) {
...
-      Node victim = policy.findVictim(sentinel, policyStats);
-      policyStats.recordEviction();
-
-      boolean admit = admittor.admit(candidate.key, victim.key);
-      if (admit) {
-        evictEntry(victim);
-      } else {
-        evictEntry(candidate);
+    if (currentSize > maximumSize) {
+      while (currentSize > maximumSize) {
```
ben-manes: If the candidate is evicted, but further evictions are needed, then you may re-evict the candidate. When doing so you decrease the current size multiple times, but only removed once. This means that the state is corrupted and no longer bounded as expected.

I believe in your case you would need to evict only the victims once the candidate was discarded.

ben-manes:
An easy way to verify the weighted LRU is to compare against Guava, if you set `concurrencyLevel(1)` in `GuavaPolicy` to ensure only a single LRU segment.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
ohadeytan: In many traces (including the provided trace of adapt-size) there are few objects with the same key but a different weight (I think the source of it is write operation in between). Those are different objects, so we should handle them as such (other implementations do so as well).

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
ohadeytan: I tried it, but it has (many?) collisions.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
ohadeytan: When the candidate is evicted we guaranteed that no more evictions needed, since the candidate weight is the reason we exceed the maximumSize in the first place. Aren't we?

ben-manes: you're right, I was thinking about W-TinyLFU where this is delayed a bit.

ben-manes:
please run `./travis.sh analysis` to fix the PMD warning

ohadeytan:
I wonder if we should move the `key = hashCode(key, weight)` rehashing to the trace parser or even to the `AccessEvent` constructor. Otherwise, each policy (e.g. Guava) would need to it in order to get the same results (I just tried it, and with the rehashing we get identical result).

ben-manes:
That's a good idea, it should be abstracted away from the policy.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/linked/LinkedPolicy.java
```diff
@@ -40,12 +44,13 @@
...
  * @author ben.manes@gmail.com (Ben Manes)
  */
-public final class LinkedPolicy implements KeyOnlyPolicy {
+public final class LinkedPolicy implements Policy {
   final Long2ObjectMap<Node> data;
   final PolicyStats policyStats;
   final EvictionPolicy policy;
   final Admittor admittor;
   final int maximumSize;
+  int currentSize;
```
ben-manes: nit: please put `currentSize` after the final fields
explanation: A nice style in Java is to declare fields as,
```
{statics}
<break>
{final instance fields}
<break>
{mutable instance fields}
```

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/product/CaffeinePolicy.java
```diff
@@ -67,8 +67,10 @@ public void record(AccessEvent event) {
     if (value == null) {
       cache.put(event.key(), event);
       policyStats.recordMiss();
+      policyStats.recordWeightedMiss(event.weight());
```
ben-manes: Would it make sense to have `recordWeightedMiss` also perform `recordMiss`, so implementations use one or the other? 

Otherwise I think these would be named `recordHitWeight` and `recordMissWeight` to indicate that they maintain the weights and not also the hit/miss counts.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/PolicyStats.java
```diff
@@ -148,11 +170,22 @@ public double hitRate() {
...
     long requestCount = requestCount();
     return (requestCount == 0) ? 0.0 : (double) missCount / requestCount;
   }
 
+  public double weightedMissRate() {
+    long requestsWeight = requestsWeight();
+    return (requestsWeight == 0) ? 1.0 : (double) missesWeight / requestsWeight;
+  }
+
+
```
ben-manes: nit: please remove extra newline

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/product/CaffeinePolicy.java
ohadeytan: Good point. I'll combine them.

PR #379, opened 2019-12-17
Add hit/miss penalty support to the simulator

ben-manes:
@himelbrand

I reduced this down to 13 files to add or modify, including adding penalties to `ClairvoyantPolicy` which you accidentally missed. This is a minimal change.

I didn't understand what `AFS` meant in the policy stats, so I didn't include it. Can you explain and maybe we'll add it back? I was concerned since it required retaining all keys which seemed odd.

I didn't add support to the `Admittor` because an implementation using penalties does not exist yet. We can add it separately, when you have an implementation to explore.


himelbrand:
Thanks,
AFS = after first seen
actually, when I think about it now, I'm not sure if it's an interesting metric, but when I ran experiments I also checked this, in order to see differences after the cache "stabilize" after initialization.
If you don't feel it's needed, it's okay, if I see I need some metric of the sort I'll talk to you about adding it at a later time. 
Regarding the Admittor I'll make the extension when I'll implement my admission policy, and I'll add it then.

ben-manes:
@himelbrand, @ohadeytan, @gilga1983 
A recent paper you might find interesting is [CHOPT](https://geraldleizhang.com/publications/CHOPT.pdf) for offline analysis which incorporates asymmetric read and write costs in the memory hierarchy. The paper is listed in the author's public publications, even though it says anonymous in the pdf.

PR #394, opened 2020-02-09
Fix SCM connection URLs when publishing

sschuberth:
Both URLs were missing the [provider] part (here "git"), see https://maven.apache.org/pom.html#SCM.
Also, the connection URL must be usable for anonymous clones, so it should not contain a user name.
Finally, GitHub does not endorse the use of the "git" protocol, so use "ssh" for the authenticated developer
connection instead.


ben-manes:
Thanks!

ben-manes:
Released

PR #403, opened 2020-04-02
A reproducer of malfunctioning "expiry after write" of the AsyncCache

stepanv:
this is a reproducer, about to create an issue for it

### DIFF
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsyncCacheExpiryReproducerTest.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsyncCacheExpiryReproducerTest.java
@@ -1,0 +1,129 @@
+package com.github.benmanes.caffeine.cache;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * A reproducer showing that {@link AsyncCache} is not always expiring correctly
+ * with {@link Caffeine#expireAfterWrite(Duration)} and hence may return stale values or load to often even if
+ * not necessary.
+ */
+@Test(singleThreaded = true)
+public final class AsyncCacheExpiryReproducerTest {
+
+    int count = 100000;
+
+    @Test
+    public void testExpiryWrite_everythingAlwaysExpire_closeToEachOther_BUG()
+            throws InterruptedException, ExecutionException, TimeoutException {
+
+        Duration expireAfterWrite = Duration.ofMillis(100);
+        long tickerIncrementGetMillis = 101;
+        long tickerIncrementCycleMillis = 101;
+        // the value should always expire before another access but it's not always like this!
+        long expectedCounterValue = count;
+
+        verify(expireAfterWrite, tickerIncrementGetMillis, tickerIncrementCycleMillis, expectedCounterValue);
+    }
+
+    @Test
+    public void testExpiryWrite_everythingAlwaysExpire_closeToEachOther_DAYS_BUG()
+            throws InterruptedException, ExecutionException, TimeoutException {
+
+        Duration expireAfterWrite = Duration.ofDays(100);
+        long tickerIncrementGetMillis = TimeUnit.DAYS.toMillis(101);
+        long tickerIncrementCycleMillis = TimeUnit.DAYS.toMillis(101);
+        // the value should always expire before another access but it's not always like this!
+        long expectedCounterValue = count;
+
+        verify(expireAfterWrite, tickerIncrementGetMillis, tickerIncrementCycleMillis, expectedCounterValue);
+    }
+
+    @Test
+    public void testExpiryWrite_everythingAlwaysExpire_farAwayFromEachOther_BUG()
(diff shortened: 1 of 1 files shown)


stepanv:
the results of the reproducer tests are:
```
java.lang.AssertionError: AsyncCacheExpiryReproducerTest#testExpiryWrite_everythingAlwaysExpire_closeToEachOther_BUG
...
Caused by: java.lang.AssertionError: expected [100000] but found [99961]
```
```
java.lang.AssertionError: AsyncCacheExpiryReproducerTest#testExpiryWrite_everythingAlwaysExpire_closeToEachOther_DAYS_BUG
...
Caused by: java.lang.AssertionError: expected [100000] but found [210]
```
```
java.lang.AssertionError: AsyncCacheExpiryReproducerTest#testExpiryWrite_everythingAlwaysExpire_farAwayFromEachOther_BUG
...
Caused by: java.lang.AssertionError: expected [100000] but found [99015]
```
```
java.lang.AssertionError: AsyncCacheExpiryReproducerTest#testExpiryWrite_everythingAlwaysExpire_farAwayFromEachOther_DAYS_BUG
...
Caused by: java.lang.AssertionError: expected [100000] but found [211]
```
```
java.lang.AssertionError: AsyncCacheExpiryReproducerTest#testExpiryWrite_firstValueNeverExpire_expireAfter1Minute_BUG
...
Caused by: java.lang.AssertionError: expected [1] but found [34]
```

ben-manes:
Can you provide details of the test / expectations? There are races inherent with AsyncCache, e.g. callbacks when the future completes, which cause cause test flakiness.

stepanv:
Thanks @ben-manes for the prompt response, I understand it may not be clear, but my expectations are that either
* the async cache value is loaded once and then never expires (1 reproducer test)
* or that the value always expires because I'm reading it too infrequently (4 reproducer tests)

stepanv:
I'm also using the `Ticker` but even when I really use just real world clock, the AsyncCache intermittently stumbles and doesn't load the value when it's expired.

stepanv:
But there might be some wrong assumptions of mine because the problem I'm outlining is quite severe because it seems some entries don't expire even after minutes after the expiration (but when I'm using artificial `Ticker`)

ben-manes:
> But there might be some wrong assumptions of mine because the problem I'm outlining is quite severe because it seems some entries don't expire even after minutes after the expiration (but when I'm using artificial Ticker)

Removal of expired entries occurs lazily when a maintenance cycle is triggered due activity on the cache. The expired entry would be visible for a size, but should not be returned an a miss emulated.

If you wish to have prompt expiration regardless of activity then you can define a `Caffeine.scheduler` and provide a thread to schedule on. In JDK9 there is a JVM-wide one, so I suggest using that via `Scheduler.systemScheduler()`.

ben-manes:
I took a fresh look at your tests to see if I could resolve them with minimal changes. The issues are due to implementation details that you didn't know to account for. I was able to make all but one pass by changing to the code below.

1. Due to callbacks and async maintenance, there will be asynchronous work that occurs in parallel to your loop. The loop should wait for the thread pool to be done before incrementing the time (see `awaitQuiescence`).
2. The ticker is called once when retrieving the value to see if it expired, and again when the future completes to set the start time. By incrementing for the next iteration's time before the future completes, you allow the cache entry to start aging at the next time cycle.
3. I switched to Guava's `FakeTicker` which tracks in nanoseconds. In your case of tracking in milliseconds, the time becomes saturated instead of overflowing to be negative. This causes expiration to halt and not appear in your test output, because `toNanos` is documented as overflowing to `Long.MAX_VALUE`.
4. In `testExpiryWrite_firstValueNeverExpire_expireAfter1Minute_BUG`, it increments the time by 10ms per iteration, 10 ms per ticker read (2x), and an expireAfterWrite of 1 minute over 100,000 iterations. If we disable the per-read increment then at iteration 6,000 that is 60,000ms for the entry which expires and breaks the test. The 100,000 iteraterations means it lasts for 16.5 minutes, so the test's expectations need to be corrected.

In summary, these seem to be mistakes in your tests and small oversights. They're all pretty reasonable and easily surprising, but none are actual implementation errors.

```java
private void verify(Duration expireAfterWrite, long tickerIncrementGetMillis,
        long tickerIncrementCycleMillis, long expectedCounterValue)
        throws InterruptedException, ExecutionException,
        TimeoutException {

    AtomicLong counter = new AtomicLong(0);
    FakeTicker ticker = new FakeTicker()
        .setAutoIncrementStep(tickerIncrementGetMillis, MILLISECONDS);;
    AsyncCache<Integer, String> target = Caffeine.newBuilder()
            .expireAfterWrite(expireAfterWrite)
            .ticker(ticker::read)
            .buildAsync();

    for (int i = 0; i < count; i++) {
        CompletableFuture<String> stringCompletableFuture = target.get(2,
                (integer, executor) -> CompletableFuture
                        .supplyAsync(() -> "hello-from-loader-" + counter.getAndIncrement()));

        System.out.printf("Iteration: %d, value: %s%n", i, stringCompletableFuture.get(10, SECONDS));
        Assert.assertTrue(ForkJoinPool.commonPool().awaitQuiescence(10, SECONDS));

        ticker.advance(tickerIncrementCycleMillis, TimeUnit.MILLISECONDS);
    }

    Assert.assertEquals(counter.get(), expectedCounterValue);
}
```

stepanv:
yeah, the last failing test is indeed my dumb oversight. Thanks a lot for your time to extend and fix the tests, it really helps to understand a piece of the iceberg that's underneath the surface of Caffeine. To explain my original motivation to write such tests, I was getting intermittent failures of my tests where we're using your cache. You explained it all way beyond what I could ask for. Thanks!!

PR #416, opened 2020-05-04
Fix comment

imangry:
The comment at `com.github.benmanes.caffeine.cache.Node#getVariableTime` is wrong.


ben-manes:
nice catch!

PR #419, opened 2020-05-13
Allow rewriting ARC traces to AdaptSize format

ohadeytan:
Change ARC parser to support size-aware events. Size is always 512.
Support rewriting of size-aware events.
Add output format of AdaptSize.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/arc/ArcTraceReader.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/arc/ArcTraceReader.java
@@ -15,11 +15,17 @@
  */
 package com.github.benmanes.caffeine.cache.simulator.parser.arc;
 
+import static com.github.benmanes.caffeine.cache.simulator.policy.Policy.Characteristic.WEIGHTED;
+
 import java.io.IOException;
+import java.util.Set;
 import java.util.stream.LongStream;
+import java.util.stream.Stream;
 
 import com.github.benmanes.caffeine.cache.simulator.parser.TextTraceReader;
-import com.github.benmanes.caffeine.cache.simulator.parser.TraceReader.KeyOnlyTraceReader;
+import com.github.benmanes.caffeine.cache.simulator.policy.AccessEvent;
+import com.github.benmanes.caffeine.cache.simulator.policy.Policy.Characteristic;
+import com.google.common.collect.Sets;
 
 /**
  * A reader for the trace files provided by the authors of the ARC algorithm. See
@@ -27,19 +33,24 @@ import com.github.benmanes.caffeine.cache.simulator.parser.TraceReader.KeyOnlyTr
  *
  * @author ben.manes@gmail.com (Ben Manes)
  */
-public final class ArcTraceReader extends TextTraceReader implements KeyOnlyTraceReader {
+public final class ArcTraceReader extends TextTraceReader {
 
   public ArcTraceReader(String filePath) {
     super(filePath);
   }
 
   @Override
-  public LongStream keys() throws IOException {
-    return lines().flatMapToLong(line -> {
+  public Set<Characteristic> characteristics() {
+    return Sets.immutableEnumSet(WEIGHTED);
+  }  
+
+  @Override
+  public Stream<AccessEvent> events() throws IOException {
+    return lines().flatMap(line -> {
       String[] array = line.split(" ", 3);
       long startBlock = Long.parseLong(array[0]);
       int sequence = Integer.parseInt(array[1]);
-      return LongStream.range(startBlock, startBlock + sequence);
+      return LongStream.range(startBlock, startBlock + sequence).mapToObj(key -> AccessEvent.forKeyAndWeight(key, 512));
     });
   }
 }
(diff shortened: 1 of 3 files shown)


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/arc/ArcTraceReader.java
```diff
@@ -15,31 +15,42 @@
...
+  }  
+
+  @Override
+  public Stream<AccessEvent> events() throws IOException {
+    return lines().flatMap(line -> {
       String[] array = line.split(" ", 3);
       long startBlock = Long.parseLong(array[0]);
       int sequence = Integer.parseInt(array[1]);
-      return LongStream.range(startBlock, startBlock + sequence);
+      return LongStream.range(startBlock, startBlock + sequence).mapToObj(key -> AccessEvent.forKeyAndWeight(key, 512));
```
ben-manes, line 53: Please keep to 100c line length, so put `.mapToObj` on the next line (similar in other changes, too)

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/rewrite/OutputFormat.java
```diff
@@ -18,18 +18,27 @@
...
-      writer.write(Long.toString(event));
+    @Override public void write(BufferedWriter writer, AccessEvent event) throws IOException {
+      writer.write(Long.toString(event.key()));
+      writer.newLine();
+    }
+  },
+  ADAPT_SIZE {
+    long time = 0;
+    @Override public void write(BufferedWriter writer, AccessEvent event) throws IOException {
+      writer.write(Long.toString(time++) + " " + Long.toString(event.key()) + " " + Integer.toString(event.weight()));
```
ben-manes, line 38: another nit.. :)

To avoid the extra string concat just to throw it away, append as multiple statements.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/rewrite/OutputFormat.java
```diff
@@ -18,18 +18,27 @@
...
   CLIMB {
-    @Override public void write(BufferedWriter writer, long event) throws IOException {
-      writer.write(Long.toString(event));
+    @Override public void write(BufferedWriter writer, AccessEvent event) throws IOException {
+      writer.write(Long.toString(event.key()));
+      writer.newLine();
+    }
+  },
+  ADAPT_SIZE {
+    long time = 0;
```
ben-manes, line 36: This is a fair pain, because you want the state but it is implicitly static which is quite ugly to modify. Adding it as a parameter would be preferable.

ben-manes:
If interested, I had a deeper refactoring that I never completed, when running into some of the same limitations. More just an fyi of where we could generalize this into a better tool.

[traces.patch.txt](https://github.com/ben-manes/caffeine/files/4620621/traces.patch.txt)

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/rewrite/OutputFormat.java
ohadeytan: What do you mean by "as a parameter"? Since AccessEvent does not have a time field the time can only be synthetic.

ben-manes: To the `OutputFormat.write` method and keep the state in the `Rewriter`.  An enum should be stateless as it represents a constant and implicitly static.

ben-manes:
squashed and merged

ben-manes:
hmm, is this giving you what you want? For a weighted trace you might consider using the starting block as the key and the sequence * block size as the weight. It wouldn’t represent a disk page cache as well, but at least adapt it to show a more varied set of  workloads for weighted policies. Given the lack of direct traces its probably a good approximation but so both the current and adjusted trace readers could be implemented.

ohadeytan:
For now, it is what I want. My intent is to run those traces with other simulators (LHD, AdaptSize) and show that they lack on unique sized traces, hence less generic cache solution. We might use your suggestion in the future.

ben-manes:
Oh good. For LHD’s simulator you might need to revert your weight + key hash change since that skews it’s results when comparing our LRU to theirs. I think you needed it for comparing against AdaptSize’s simulator only.

ohadeytan:
Why it should do any difference?

ben-manes:
Forwarded the thread, but it was different enough to be visible when comparing LHD and our simulator for a baseline LRU.

ben-manes:
We overlooked that the changes to `ArcTraceReader` implicitly filters the downstream policies to be weight-based. I'll revert that file, as I think you'd be better off by a local hack to the adapt size writer to an explicit weight, since you only need to translate the files once.

ohadeytan:
Hmmm. I thought that key only policies can use weight-aware parsers and just ignore the weight. Isn't it more reasonable?

ben-manes:
That would make the report confusing as it would mix weight and key-based policies together. The key-based would have a hidden advantage, as it would allow far more entries to reach the maximum. It would be fine if we could better distinguish them, e.g. output groups into different tables, so that it was much more explicit.

ohadeytan:
Make sense. As you said, it's only a one time gig for now, so you can revert it and maybe we would expand it in the future.

PR #421, opened 2020-05-19
Make the package discoverable in README.md

alexanderkjeldaas:


ben-manes:
Can you provide a little more context please? Most developers work from an IDE which would hande discovering the import trivially.

alexanderkjeldaas:
Ah, the reason for this PR is that I was looking for a self-contained example and it took a while to track down the code.  I prefer self-contained examples.

ben-manes:
well, technically it is not, either. There is no `Key`, `Graph`, `LoadingCache`, etc imports. So it would still be only a psuedo sample.

There is an `examples` directory, though we haven't built it out much. If you're interested that that would be a place to show runnable samples or tutorial like walkthrough.

alexanderkjeldaas:
Feel free to close this PR.

PR #423, opened 2020-05-26
Update AsyncCache.java docs

alexanderkjeldaas:
Document the difference between returning *the* cached future, or *a* future associated with a key.

```
CompletableFuture f = CompletableFuture.completedFuture(v);
cache.put(k, f);
assert(cache.getIfPresent(k).isDone());  // holds
assert(cache.get(k, (k) => v).isDone()); // does NOT hold / undefined
```


ben-manes:
I don’t think that I follow. We don’t wrap or copy a given future, and instead act as a bounded `Map<K, CompletableFuture<V>>`. Your call to `get` should work assuming there are no concurrent modifications.

If you mean that as two independent examples, then of course resolving a future is asynchronous by its definition. An immediate future can be returned by a mapping function. In your case it is automatically wrapping the function and inserting that future into the cache, which introduces a slight race. A test might use a same thread executor to disable any asynchronous computations for simplicity, resolving that race.

alexanderkjeldaas:
I guess I have some other issue in my tests then.. closing this.

ben-manes:
For tests, consider using `Caffeine.executor(Runnable::run)` unless you want to assert asynchronous behavior.

PR #448, opened 2020-08-18
fix #447: Allow the CaffeineCachingProvider to be used with OSGi DS

### LINKED ISSUE #447: Allow the CaffeineCachingProvider to be used with OSGi DS
Although the `jcache` module is an OSGi bundle, the `CaffeineCachingProvider` can only be used directly, by tying the consumer to the `com.github.benmanes.caffeine.jcache` implementation. If the `CaffeineCachingProvider` would be an OSGi `Component`, a consumer would only depend on the `javax.cache` API and the `CaffeineCachingProvider` would be an opaque implementation.

Ehcache has a similar approach for their JSR-107 implementation.

raducotescu:
* annotated CaffeineCachingProvider to make it an OSGi service
* when running in an OSGi container, set the CaffeineCachingProvider's classloader as the Thread context classloader for every API call on the provided CacheManagerImpl - this makes sure that the default config is read from the bundle; the default classloader will do a fallback to the app classloader, allowing to override / define configurations
* added an OSGi DS test
* made the bnd build reproducible (no extra bundle headers)
(fixes #447)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,17 +73,9 @@ subprojects {
     testRuntimeOnly testLibraries.osgiRuntime
   }
 
-  configurations {
-    bundleCompile
-  }
-
-  sourceSets {
-    bundle
-  }
-
   task bundle(type: aQute.bnd.gradle.Bundle) {
-    from sourceSets.bundle.output
-    sourceSet = sourceSets.bundle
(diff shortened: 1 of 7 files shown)


ben-manes:
Thank you!

Issue #446, opened 2020-08-18
Unsafe Access to Volatile `accessTime` does not use `Unsafe.{get|put}LongVolatile`


jamesdbowman:
Hi,

My colleague @caoman has discovered that the generated code to get/put `accessTime` does not properly use the `Unsafe.{get|put}LongVolatile` methods.

Would it be possible to pass the volatility of a given field to `newGetter` and `newSetter` methods such that the proper `Unsafe.*Volatile` method could be used for interaction? E.g. [here](https://github.com/ben-manes/caffeine/blob/801e532935dd822aab75e9625a520df7b62bb0b0/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/node/AddExpiration.java#L117-L120) for `accessTime`?

Thanks in advance for your thoughts,
James

ben-manes:
I think the expectation was that it would piggyback on other memory barriers, so strictly wasn't necessary. Was an issue observed?

jamesdbowman:
[Java TSAN](https://wiki.openjdk.java.net/display/tsan/Main) found a (theoretical) data race between one thread calling `BoundedLocalCache.setAccessTime` and another thread (running the cleanup task) calling `BoundedLocalCache.expireAfterAccessEntries` (getting the `accessTime` value).

ben-manes:
right, that makes sense. It is benign because if a stale read views it as expired, the call to `evictEntry` will validate within the computation. That read piggybacks on the lock's barrier to have a updated read and [resurrects](https://github.com/ben-manes/caffeine/blob/ecd5e13e1b8abe9796e0ec5c9ac30aa691f7d472/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L913-L927) the entry.

jamesdbowman:
Ok, thanks for the clarification Ben.

Then it does seem that the absence of `Unsafe.*Volatile` methods is unrelated, that such races are inherently built into the concurrency model of the library, and that running Java TSAN on this library will find false positives.

Closing for now, thanks for taking a look.

caoman:
I not convinced this is benign. The bigger issue is the atomicity violation on a long variable. [JLS 17.7 Non-Atomic Treatment of double and long](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.7) says "a single write to a non-volatile long or double value is treated as two separate writes", and "Writes and reads of volatile long and double values are always atomic." 

[...]

PR #449, opened 2020-08-18
Add an optional `Volatility` argument to `NodeRule#new{Getter|Setter}`.

jamesdbowman:
By default, `Volatility.NON_VOLATILE` is used.

Tested by manually inspecting the generated sources.

#446

### DIFF
--- a/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/node/AddExpiration.java
+++ b/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/node/AddExpiration.java
@@ -86,14 +86,14 @@ public final class AddExpiration extends NodeRule {
   private void addVariableTime(String varName) {
     MethodSpec getter = MethodSpec.methodBuilder("getVariableTime")
         .addModifiers(Modifier.PUBLIC)
-        .addStatement("return $T.UNSAFE.getLong(this, $N)",
+        .addStatement("return $T.UNSAFE.getLongVolatile(this, $N)",
             UNSAFE_ACCESS, offsetName(varName))
         .returns(long.class)
         .build();
     MethodSpec setter = MethodSpec.methodBuilder("setVariableTime")
         .addModifiers(Modifier.PUBLIC)
         .addParameter(long.class, varName)
-        .addStatement("$T.UNSAFE.putLong(this, $N, $N)",
+        .addStatement("$T.UNSAFE.putLongVolatile(this, $N, $N)",
             UNSAFE_ACCESS, offsetName(varName), varName)
         .build();
     MethodSpec cas = MethodSpec.methodBuilder("casVariableTime")
@@ -116,8 +116,10 @@ public final class AddExpiration extends NodeRule {
     }
     context.nodeSubtype.addField(newFieldOffset(context.className, "accessTime"))
         .addField(long.class, "accessTime", Modifier.VOLATILE)
-        .addMethod(newGetter(Strength.STRONG, TypeName.LONG, "accessTime", Visibility.LAZY))
-        .addMethod(newSetter(TypeName.LONG, "accessTime", Visibility.LAZY));
+        .addMethod(
+            newGetter(Strength.STRONG, TypeName.LONG, "accessTime", Visibility.LAZY,
+                Volatility.VOLATILE))
+        .addMethod(newSetter(TypeName.LONG, "accessTime", Visibility.LAZY, Volatility.VOLATILE));
     addTimeConstructorAssignment(context.constructorByKey, "accessTime");
     addTimeConstructorAssignment(context.constructorByKeyRef, "accessTime");
   }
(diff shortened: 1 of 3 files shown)


ben-manes:
Running the [GetPutBenchmark](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/GetPutBenchmark.java) and it appears that while reads are unaffected (as volatile is mostly free), writes do degrade. That's hard to pinpoint if noise (JIT, background work) as it maximizes the cpu either way, but over a few runs it does seem a bit slower. That's still an acceptable performance, but an fyi. I changed to run with `expireAfterWrite` and setting the `ticker` to a constant value to avoid nanoTime overheads.

The argument for making the value a volatile is weaker than the expiration timestamps (a reference cannot see any tearing). The existing logic mirrors Guava's [Suppliers.memoize](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Suppliers.java#L112-L117) which piggybacks on the fence from a preceding volatile read and publishing on the unlock. That code, as well as optimizations throughout j.u.c., were written by Martin Buchholz who should still be in your Java core team at Google for reference. While I understand the spec obsession when running a sanitizer, one does not still write C towards the PDP-11. Allowing benign races which can do no harm is common in concurrency code, especially when optimizing against memory fences and HotSpot directly.

If you have a clean environment to benchmark on (rather than my macbook!), I think that would make me more comfortable merging this in. I think any degradation should be acceptable, modest, and only in a synthetic stress test so as to have no real world implications. That, or Martin telling me to just let it though, would avoid my unease as otherwise I feel that I'd do so merely to satisfy a pedantic sanitizer.

jamesdbowman:
### Summary

A benchmark on six-core workstation showed the same slowdown in writes/sec.

I am definitely not well-versed in JVM nor Caffeine internals to be able to make a convincing argument in one direction or the other; I am merely a happy consumer of this library with a thread sanitization violation filed against my team. I feel inclined to agree that the change only satisfies a pedantic sanitizer for measurable performance loss and doesn't seem ideal. Though, again, I don't know enough about JVM internals to make such an argument!

To paraphrase Martin Buchholz on the Java team: he seems to agree that using non-volatile `Unsafe` methods to access volatile fields ("piggy-backing" as it has been referred to) is a standard access pattern, and that this change is unnecessary. I can't find a GitHub account for Martin, or I'd tag him here.

@caoman, any additional thoughts on this change?

### Benchmark Environment

```
$ lscpu
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   46 bits physical, 48 bits virtual
CPU(s):                          12
On-line CPU(s) list:             0-11
Thread(s) per core:              2
Core(s) per socket:              6
Socket(s):                       1
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           85
Model name:                      Intel(R) Xeon(R) W-2135 CPU @ 3.70GHz

```

### Benchmark Results

Without change:

```
# Threads: 8 threads (1 group; 6x "readwrite_get", 2x "readwrite_put" in each group), will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite
# Parameters: (cacheType = Caffeine)
```
```
Result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite":
  153748914.778 ±(99.9%) 19488508.653 ops/s [Average]
  (min, avg, max) = (133638324.057, 153748914.778, 178772120.076), stdev = 12890441.822
  CI (99.9%): [134260406.125, 173237423.430] (assumes normal distribution)

Secondary result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite:readwrite_get":
  135861653.970 ±(99.9%) 20283023.196 ops/s [Average]
  (min, avg, max) = (115874953.723, 135861653.970, 162774214.374), stdev = 13415963.999
  CI (99.9%): [115578630.774, 156144677.166] (assumes normal distribution)

Secondary result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite:readwrite_put":
  17887260.808 ±(99.9%) 1421922.979 ops/s [Average]
  (min, avg, max) = (15997905.702, 17887260.808, 18992842.465), stdev = 940514.011
  CI (99.9%): [16465337.828, 19309183.787] (assumes normal distribution)
```

---

With change:

```
# Threads: 8 threads (1 group; 6x "readwrite_get", 2x "readwrite_put" in each group), will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite
# Parameters: (cacheType = Caffeine)
```
```
Result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite":
  160470832.933 ±(99.9%) 11953518.502 ops/s [Average]
  (min, avg, max) = (141874473.204, 160470832.933, 174325726.558), stdev = 7906512.374
  CI (99.9%): [148517314.431, 172424351.435] (assumes normal distribution)

Secondary result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite:readwrite_get":
  143351054.698 ±(99.9%) 12311279.297 ops/s [Average]
  (min, avg, max) = (124467659.876, 143351054.698, 158301622.831), stdev = 8143148.988
  CI (99.9%): [131039775.402, 155662333.995] (assumes normal distribution)

Secondary result "com.github.benmanes.caffeine.cache.GetPutBenchmark.readwrite:readwrite_put":
  17119778.235 ±(99.9%) 970339.869 ops/s [Average]
  (min, avg, max) = (15984815.749, 17119778.235, 17809465.075), stdev = 641819.743
  CI (99.9%): [16149438.366, 18090118.103] (assumes normal distribution)
```

caoman:
I understand the concern. Let's do more discussion internally (mainly to get Martin's response), then follow up here.

ben-manes:
Closing due to lack of feedback. This would be done slightly differently in the v3 branch where we have replaced Unsafe with VarHandles. Thanks for the code and discussions, though.

@caoman please reopen if you have feedback and would like to revisit.

PR #457, opened 2020-09-25
BoundedLocalCache Refactoring for Eviction abstractions

automenta:
Abstracts the eviction candidate/victim heuristic.  it's still hardcoded in this commit with a "sketch frequency divided by node weight" experiment - but it's where a new interface or extension to Weigher could support custom impls provided in CacheBuilder.

Also :boom: Remapping and Eviction inner classes replace some Map-method lambdas that relied on many 1-element-array local variables.

Other misc refactoring that a line-by-line diff should explain

:warning: Unit tests *seem* to be intact.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
@@ -340,6 +340,10 @@ abstract class AbstractLinkedDeque<E> extends AbstractCollection<E> implements L
     return removeFirst();
   }
 
+  public final PeekingIterator<E> iterator(boolean ascending) {
+    return ascending ? iterator() : descendingIterator();
+  }
+
   @Override
   public PeekingIterator<E> iterator() {
     return new AbstractLinkedIterator(first) {
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
@@ -15,16 +15,14 @@
  */
 package com.github.benmanes.caffeine.cache;
 
-import static com.github.benmanes.caffeine.cache.Async.ASYNC_EXPIRY;
-import static com.github.benmanes.caffeine.cache.Caffeine.ceilingPowerOfTwo;
-import static com.github.benmanes.caffeine.cache.Caffeine.requireArgument;
-import static com.github.benmanes.caffeine.cache.Caffeine.requireState;
-import static com.github.benmanes.caffeine.cache.LocalLoadingCache.newBulkMappingFunction;
-import static com.github.benmanes.caffeine.cache.LocalLoadingCache.newMappingFunction;
-import static com.github.benmanes.caffeine.cache.Node.PROBATION;
-import static com.github.benmanes.caffeine.cache.Node.PROTECTED;
-import static com.github.benmanes.caffeine.cache.Node.WINDOW;
-import static java.util.Objects.requireNonNull;
+import com.github.benmanes.caffeine.base.UnsafeAccess;
+import com.github.benmanes.caffeine.cache.Async.AsyncExpiry;
+import com.github.benmanes.caffeine.cache.LinkedDeque.PeekingIterator;
+import com.github.benmanes.caffeine.cache.References.InternalReference;
+import com.github.benmanes.caffeine.cache.stats.StatsCounter;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.io.InvalidObjectException;
 import java.io.ObjectInputStream;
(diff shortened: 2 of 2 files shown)


ben-manes:
Thanks, but I’m not sure what the goal is. Why would we want to abstract the eviction candidate/victim heuristic?

If this is to later evolve to support alternative eviction policies, that’s something we’d design in only after implementing in the simulator to fully evaluate the choice.

I’m afraid this is was premature work because I don’t yet know the benefits.

automenta:
Is there one universally optimal eviction policy for every possible Caffeine use-case?  If not then users should be able to evaluate their own policies at runtime.  Especially policies which are not only a function of estimated frequency, but also of the data itself, which cant be simulated ahead of time.

ben-manes:
The current policy is fairly robust at being the at or competitive to the top algorithm. Since a policy requires its own metadata, e.g. many use ghost lists, it is very hard to abstract out efficiently. Those details do impact the internals to a degree that makes it impractical to be fully dynamic. Instead we've done research to improve the algorithms and keep the algorithm details hidden from the apis.

The simulator is where we add new policies to evaluate against traces. If there are demonstrated improvements then we consider how to integrate them into the library. If you have ideas then adding to the simulator would let us compare more thoroughly and have a quantifiable idea of the benefits when integregrating into the library.

@ohadeytan has been working on improving the eviction policy to take into account the entry's weight. That appears to be your goal here, too. In that case if you send us an email then I think he'd be happy to discuss his findings and share the pre-print paper. Once that work is public, we'd welcome help to implement the changes into the library.

automenta:
its pretty clear how some kind of abstract Evictor interface can replace the hardcoded policy methods in BoundedLocalCache with negligible overhead.

in my use case, access frequency alone does not necessarily represent the total "value" of holding a cached item.  im sure there are other completely different applications where it's also the case.  

everything else in Caffeine works great.  no need to share the top secret preprint - ill wait for the publication and see how i can help then.

ben-manes:
Okay. You are welcome to send a PR to add a new policy to the simulator. If you have a trace showing benefits of your ideas we can add it to our gauntlet.

ben-manes:
I'll close this for now and ping you when the paper is public, so that we can start discussing whether and how to incorporate its improvements.

automenta:
thanks.  one quick question

https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L1385

```
          casDrainStatus(IDLE, REQUIRED);
          scheduleDrainBuffers();
```
should this be:
```
          if (casDrainStatus(IDLE, REQUIRED))
             scheduleDrainBuffers();
```
?

ben-manes:
It could. There doesn’t appear to be any harm though because it’s work is in the buffer (so will be drained) and the `scheduleDrainBuffers` has its own guards when called concurrently. I think it’s benign?

automenta:
i dont know, i found it while trying to understand the cause of the maintenance procedures.    the 'if' version will of course invoke 'scheduleDrainBuffers' potentially less frequently which may be what is intended, ex: in cases where it's already scheduled.  

otherwise there's no point in using CAS here, just setVolatile or something will make the unconditional need to schedule clear.

fwiw i also ended up changing removal notify to execute inline (ie. not through the executor) since when it happens frequently it creates lots of tasks.  but in my app i can handle the removal notification on the same thread so having to process these asynchronously was wasteful.  somehow discovering this is related to the above question.

ben-manes:
The CAS is desired because it may be in a processing state, so we don’t want to revert that and cause it to tryLock. When a CAS fails it now has a new read of the variable so the check in the schedule method will early return.

Did you use `Caffeine.executor(Runnable::run)` to disable or fork the code?

automenta:
yes originally i was using DirectExecutor, but in evaluating async maintenance i decided that at least the removal notification process was ideally non-async when removal rate is high.

i imagine: one CAS AtomicBoolean/VarHandle volatile boolean can be used to ensure at most 1 maintenance task is ever scheduled (and immediately cleared when it begins executing some time later), and a separate volatile int bitset indicating which subtasks it needs to execute when it does -  in the same maintenance task instance, or separately decomposed if requested by a parameter.  processing a queue of removal notifications in a batch would be one such subtask.  a single maintenance task instance could be safely  recycled this way too.

ben-manes:
The drainStatus state machine ensures that we have one running / scheduled instance. The CAS and tryLock lets us cheaply handle that. When it is running it starts with the state `PROCESSING_TO_IDLE` and can transition externally to `PROCESSING_TO_REQUIRED`. This way when the maintenance completes it can go to either `IDLE` or `REQUIRED`, e.g. a write occurred during its execution so a new run may be needed for eviction.

The maintenance steps are very cheap because all data structures are amortized O(1). The no-op case is instant so we haven't needed the bitset idea to mark subtasks. Most often it is just polling from a ring buffer. The most likely place it can slow down when removing the entry from the map (via a `computeIfPresent`) because `ConcurrentHashMap` uses bin locking so it would wait if colliding with a long-running user computation. The maintenance task and removal listener are async by default to offer more predictable latencies, but generally work fine if a direct executor. The maintenance task instance is recycled.

For the removal notifications we submit them directly to the executor. Since Guava lacks a executor, we pushed them to a queue and polled to execute outside of any locks. If the listener should be async from callers we expected users to make the async submission inside the listener (e.g. `RemovalListeners#asynchronous(listener, executor)`. That decoupling is nice assurance, but requires all callers to aggressively poll from this CLQ which is a small hotspot. Caffeine submits directly to the executor, which has the flaw where a direct executor could run the task under the maintenance lock. Since the ring buffers and state machines allows for a lot of concurrency, this hasn't been a problem since even a long-running execution wouldn't block other threads unless it hits a write threshold (which then applies backpressure to ensure no runaway growth). We assume that the Executor will properly handle batching (e.g. per-thread run queues).

I think we have it pretty well optimized, but if you have observed any issues and have ideas for improving that then let me know. A test case showing a bottleneck would be an important first step if we think there are aspects to improve.

PR #468, opened 2020-11-02
Update ClockPro based on author's paper

chanyoung:
@ben-manes I've sent an e-mail to request a review about this PR.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
@@ -15,21 +15,20 @@
  */
 package com.github.benmanes.caffeine.cache.simulator.policy.irr;
 
-import static com.google.common.base.Preconditions.checkState;
-import static java.util.Objects.requireNonNull;
-
-import java.util.Set;
-
 import com.github.benmanes.caffeine.cache.simulator.BasicSettings;
 import com.github.benmanes.caffeine.cache.simulator.policy.Policy;
 import com.github.benmanes.caffeine.cache.simulator.policy.Policy.KeyOnlyPolicy;
 import com.github.benmanes.caffeine.cache.simulator.policy.PolicyStats;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.Ints;
 import com.typesafe.config.Config;
-
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkState;
 
 /**
  * The ClockPro algorithm. This algorithm differs from LIRS by replacing the LRU stacks with Clock
@@ -45,22 +44,21 @@ import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
  * Effective Improvement of the CLOCK Replacement</a> and
  * <a href="http://www.slideshare.net/huliang64/clockpro">Clock-Pro: An Effective Replacement in OS
  * Kernel</a>.
- * <p>
- * This implementation is based on
- * <a href="https://bitbucket.org/SamiLehtinen/pyclockpro">PyClockPro</a> by Sami Lehtinen,
- * available under the MIT license.
  *
  * @author ben.manes@gmail.com (Ben Manes)
+ * @author park910113@gmail.com (Chanyoung Park)
  */
 public final class ClockProPolicy implements KeyOnlyPolicy {
   private final Long2ObjectMap<Node> data;
   private final PolicyStats policyStats;
 
+  // We place all the accessed pages, either hot or cold, into one single list in the order of their
+  // accesses. In the list, the pages with small recency are at the list head, and the pages with
+  // large recency are at the list tail.
+  private Node listHead;
+
   // Points to the hot page with the largest recency. The position of this hand actually serves as a
-  // threshold of being a hot page. Any hot pages swept by the hand turn into cold ones. For the
-  // convenience of the presentation, we call the page pointed to by HAND as the tail of the list,
-  // and the page immediately after the tail page in the clockwise direction as the head of the
-  // list.
+  // threshold of being a hot page. Any hot pages swept by the hand turn into cold ones.
   private Node handHot;
 
   // Points to the last resident cold page (i.e., the farthest one to the list head). Because we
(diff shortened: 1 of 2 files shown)


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
```diff
@@ -73,263 +70,556 @@
...
-    } else {
+      node = new Node(key);
+      data.put(key, node);
+      onMiss(node);
+    } else if (node.isResident()) {
       onHit(node);
+    } else {
+      onMiss(node);
+    }
+    organizeHands();
```
ben-manes, line 148: Does this violate the benefits of Clock for read concurrency?

The main benefit of Clock algorithms is to not modify central, shared state on reads by instead only updating per-entry metadata. This then leads to a higher cost on write when choose a victim. However this method updates the hands, e.g. it leads to `terminateTestPeriod` which modifies the linked lists. This can't be done safely or efficiently outside of a lock, so wouldn't it defeat the gains compared to using LIRS directly?

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
```diff
@@ -73,263 +70,556 @@
...
+  }
 
-    if (handHot == null) {
-      handHot = handCold = handTest = node;
-      node.next = node.prev = node;
+  private void onNonResidentFullMiss(Node node) {
+    // If the cold page is in the list, the faulted page turns into a hot page and is placed at the
+    // head of the list. We run handHot to turn a hot page with a large recency into a cold page.
+    //
+    // While not described in the paper, the author's reference implementation not promote the
```
ben-manes, line 207: > reference implementation not promote the

`does not`

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
```diff
@@ -45,22 +44,20 @@
  * Effective Improvement of the CLOCK Replacement</a> and
  * <a href="http://www.slideshare.net/huliang64/clockpro">Clock-Pro: An Effective Replacement in OS
  * Kernel</a>.
- * <p>
- * This implementation is based on
- * <a href="https://bitbucket.org/SamiLehtinen/pyclockpro">PyClockPro</a> by Sami Lehtinen,
- * available under the MIT license.
  *
  * @author ben.manes@gmail.com (Ben Manes)
```
ben-manes, line 48: Please update yourself to be the author 😃

ben-manes:
This is really well done, thank you!

1. There are some minor static analysis warnings if you run `./travis.sh analysis`.
2. I think it is okay to add the author's C code to this repository, but you can check with them if concerned. It could go next to their [LIRS code](https://github.com/ben-manes/caffeine/tree/master/simulator/src/main/resources/com/github/benmanes/caffeine/cache/simulator/parser/lirs).

Afterwards these comments please rebase and we'll merge it in. From your email, the hit rate improvements are dramatic and this is a much more proper version. I'd like to run a few tests and see it in action, mostly out of awe that you were able to decipher and write this! 😺

ben-manes:
I ran some traces to confirm your improvements from the email. It's not only more correct, but much better across the board!

The corda workload (blockchain mining, lru biased) confuses LIRS such that it can't correct when transitioning to glimpse, oddly enough. The lack of adaptivity means hurts everything except for Caffeine, which is able to readjust and optimize accordingly. The ARC traces shows very competitive results. Some workloads hit an edge case causing the Python port to never complete (scarab, s3).

|Trace|Size|Lirs|PyClockPro|ClockPro|Difference|
:--|--|--|--|--|--|
| multi1 | 512 | 56.25 | 52.69 | 56.01 | +3.5%
| multi3 | 512 | 45.49 | 42.07 | 40.77 | -3.5%
| loop | 512 | 49.95 | 10.66 | 50.05 | +39.5%
| glimpse | 512 |34.33 | 16.61 | 33.60 | +17%
| corda | 512 |0.01 | 0.04 | 33.31 | +33.5%
| corda + 5x loop + corda | 512 |0.31 | 0.03 | 21.33 | +21.5%
| OLTP | 512 | 27.23 | 26.22 | 27.40 | +1%
| S3 | 400K | 34.69 | timed out | 36.65 | 
| DS1 | 4M | 38.14 | 25.31 | 39.35 | +14%

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
chanyoung: Yes, it violates the benefits of clock, it's my mistake. I moved the organizeHands() into the onMiss() method. The test results are same because onHit() never change the status of clock entries so calling organizeHands() after onHit() not cause hands moving.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
chanyoung: Done.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
chanyoung: Done. 😃

chanyoung:
> There are some minor static analysis warnings if you run ./travis.sh analysis.

Done.

> I think it is okay to add the author's C code to this repository, but you can check with them if concerned. It could go next to their LIRS code.

I'll ask to the author first and if they say ok then I'll send a separate pull request about this.

> I ran some traces to confirm your improvements from the email. It's not only more correct, but much better across the board!

Thanks Ben. In my tests, pyClockPro also shows good performance in the most of traces if it is tuned to same as the PR.

- Limits the boundaries of coldTarget to 1%, 99%
- Starts the coldTarget at the minimum resident cold size
- Hot page warmup

However, even after tuned it still shows same limitation in the traces corda and S3. I think the biggest problem of the prior is it misses some important points of author's algorithm, so I can't explain why it shows good or bad performance. I think the reuse distance concept is obscure in the prior code mainly two reasons, 1) it has different test period concept and 2) handHot is not moving to the next hot entry after demotion, so it can't make sure there are no cold page in test period have larger recency than any hot page.

ben-manes:
Very cool, I'm excited to see your work make it into Pebble! Thanks a lot for writing a Java version, as I know you're going to have to redo it a bit for the Rust code.

chanyoung:
@ben-manes Could you tell me how to run mixed workloads like "corda + 5x loop + corda"? I tried like this in the reference.conf file but I think this is not the same workload what you tested.
```
paths = [ "corda:trace_vaultservice.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "corda:trace_vaultservice.gz" ]
```

ben-manes:
That looks almost right, except that I used `corda:trace_vaultservice_large.gz`. Does it match with that change?

ben-manes:
trying again and I don't get the results, so maybe I screwed those up? I'll have to backtrack, but I might have messed up that table row.

ben-manes:
yeah, scrap that row as wrong. Sorry!

ben-manes:
I found it, I accidentally used glimpse from the prior row

```
paths = [
  "corda:trace_vaultservice_large.gz", 
  "lirs:gli.trace.gz", 
  "lirs:gli.trace.gz", 
  "lirs:gli.trace.gz", 
  "lirs:gli.trace.gz", 
  "lirs:gli.trace.gz", 
  "corda:trace_vaultservice_large.gz",
 ]
```

chanyoung:
I see, thank you for check this again. I think I need to double check the adaptivity of this algorithm against to the original research code. I'll share the result to you. Thank you again.

ben-manes:
Maybe you could use the [Rewriter](https://github.com/ben-manes/caffeine/blob/master/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/Rewriter.java) to output into the LIRS trace format and try it against their C code?

chanyoung:
Thank you for the advice. I'll do that.

chanyoung:
@ben-manes Could you give me some more mixed workload set examples which are good to test adaptivity? like corda + 5x loop + corda. I think I need to see some more test results.

ben-manes:
Unfortunately that's my best one. It's a unique case of very different, extreme workloads (LRU, MRU) that are similarly sized making it possible to chain together. Otherwise we tested individual workloads to ensure robustness ([paper](https://dl.acm.org/doi/10.1145/3274808.3274816)). When brainstorming approaches, I found some of my ideas could adapt once and get stuck. I tried mixing a few workload types ad hoc, e.g. adding oltp or web12, but corda + loop/gli is the strongest. 

Much later when implementing into Caffeine I found it useful to chart the adaptivity by manually instrumenting the cache to print out the metrics at sample points. This showed that the hill climbing logic is able to converge to a good configuration and restart when needed. Maybe you can do something similar?

| height = 400px

cruftex:
Nice to see some activity here! BTW, I did check in the code from the Clock Pro Plus paper here:
https://github.com/cache2k/cache2k-benchmark/tree/master/clockProPlus

My experiments showed that CP+ needs about 10x of the scans per eviction then I need with my augmented Clock Pro algorithm in cache2k. So I don't see it production ready. I hope to find to dive into that again beginning of next year. However, maybe you like to do more experiments with it?

ben-manes:
@chanyoung implemented ClockPro+ in #471 based on the author's paper & code into a copy of this corrected version. That way it isn't skewed by not being based on a proper implementation of ClockPro, since there are confusing subtleties that @chanyoung had to resolve by digging into ClockPro's reference C code to unravel against the original paper.

Over emails we also concluded when comparing the two adaptation schemes that CP+ wasn't as good in general by having a lower hit rate. We didn't look yet into the scan length, but it's something we talked about needing to investigate before @chanyoung starts writing a production-facing version (a Golang cache for [Pebble](https://github.com/cockroachdb/pebble).

A weakness in CP's adaption is that it can be confused by noise and converge to an non-optimal setting. The below example shows this in `corda-large` + `5x loop` + `corda-large` where the hit rate swings wildly in the last phase.

ClockPlus adaptation

@chanyoung is starting to experiment applying the larger step size, decay rate, and restart threshold approach used by Caffeine's hill climbing into ClockPro. We're hoping that might help the algorithm converge to a better setting. I haven't looked into the changes yet, but an initial draft is in this [commit](https://github.com/chanyoung/caffeine/commit/f46dc538378401be09e157a43652c701b1616ac1) (L419-471). 

LIRS and ClockPro both have a sizable gap on DS1 (db) and S3 (search) compared to Caffeine, though are still much better than alternatives. I think this might be a fundamental limit of IRR and I don't have any good ideas for how to boost the performance here. Since Pebble is the storage layer for CockroachDB they seem important workloads to spend some time exploring, though.

You might want to review @chanyoung's version of ClockPro for your cache as it does have higher hit rate in some workloads due to implementing adaptation (which yours currently lacks adaption, if the comment in `ClockProPlusEviction` is still accurate).

cruftex:
@chanyoung thanks for the work and @ben-manes thanks for the overview and update what is going on.

> which yours currently lacks adaption, if the comment in ClockProPlusEviction is still accurate

Yes.  cache2k works still with a static setting. The outcome is not perfect, but still very good for most traces.

When I get back to it, I will put updates here:
https://github.com/cache2k/cache2k/issues/101

ben-manes:
@cruftex A mixture of CP and CP+ resulted in very robust performance that resolves the above degradation. It combines both adaptation schemes to compensate for each other's weaknesses. Across all workloads that we've tried it on it stays very competitive. The only minor drawback is that in some cases where CP+ beats CP by a few points (OLTP, UMass Youtube, Umass Financial), the combined algorithm is closer to CP than CP+. That's overall not too bad, a 3-5% difference, while staying robust and converging quickly. This seems like a good win to me and @chanyoung plans on writing an analysis to solicit feedback from the original authors. We'll share that with you, of course, and that variation might be of interest for your own usage.

PR #469, opened 2020-11-02
Verify maximumWeight before setting

alllex:
The `maximumWeight` builder method is the only method that does the argument verification before the assignment.
It looks like a typo, and does not affect anyone in practice, but it is still better to have the implementation uniform.


ben-manes:
Thanks! This same typo is in Guava's CacheBuilder, if you'd like to send a PR to them too.

PR #471, opened 2020-11-09
Add ClockProPlus policy

chanyoung:
- It seems dirty, but I just copied and pasted the ClockPro code and modified parts of adjusting coldTarget. This is because the noise issue still remaining in ClockPro algorithm, so I wanted to keep the code as easy as possible to modify and test without depending on other files.
- ClockPro's default configuration has been changed as we discussed

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Ben Manes. All Rights Reserved.
+ * Copyright 2020 Ben Manes. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
```diff
@@ -0,0 +1,663 @@
...
+/**
+ * The ClockProPlus algorithm. This algorithm differs from ClockPro by adjusting the coldTarget
+ * with the utility-driven adaption idea borrowed from CAR. The algorithm is explained by the
+ * authors in <a href="https://dl.acm.org/doi/10.1145/3319647.3325838">CLOCK-Pro+: improving
+ * CLOCK-Pro cache replacement with utility-driven adaptation</a>.
+ *
+ * @author ben.manes@gmail.com (Ben Manes)
+ * @author park910113@gmail.com (Chanyoung Park)
+ */
+public final class ClockProPlusPolicy implements KeyOnlyPolicy {
```
ben-manes, line 42: It might be helpful to call out where in the code it differs, as from a quick glance it is not obvious.

ben-manes, line 42: You had concerns about the % cold differing between CP and CP+, maybe mention that?

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
```diff
@@ -0,0 +1,663 @@
...
+  // +----- Status ------+- Resident -+- In Test -+
+  // |               HOT |       TRUE |     FALSE |
+  // |          COLD_RES |       TRUE |     FALSE |
+  // |  COLD_RES_IN_TEST |       TRUE |      TRUE |
+  // |      COLD_NON_RES |      FALSE |      TRUE |
+  // |      OUT_OF_CLOCK |      FALSE |     FALSE |
+  // +-------------------+------------+-----------+
+  enum Status {
+    HOT, COLD_RES, COLD_RES_IN_TEST, COLD_NON_RES, OUT_OF_CLOCK,
+  }
```
ben-manes, line 538: fyi, no changes expected.

A nice Java-specific feature is that enums can carry both state and methods. This table could be encoded into the value, e.g.

```java
enum Status {
  HOT(/* resident */ true, /* in test */ false),
  ...

  final boolean resident;
  final boolean inTest;

  Status(boolean resident, boolean inTest) {
    this.resident = resident;
    this.inTest = inTest;
  }

  public boolean isInTest() {
    return inTest;
  }
  ...
}
```

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
```diff
@@ -310,7 +310,7 @@ private boolean runHandHot(Node trigger) {
           demoted = true;
           break;
         }
-      } else if (handHot.isInTest()) {
```
ben-manes, line 313: can you explain this to me (not in code) please? Why was this wrong and how did it impact results?

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
```diff
@@ -0,0 +1,663 @@
+/*
+ * Copyright 2015 Ben Manes. All Rights Reserved.
```
ben-manes, line 2: please update year. You can self-assign if you'd like

ben-manes (approved):
Please fix the analysis errors. In the new github actions CI, the script is `.github/scripts/analyze.sh`.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPolicy.java
chanyoung: Yes of course! In the clock, two types of entries can be exist: hot and cold. There are three types of cold entries: _"resident cold", "resident cold in test", "non resident"._ _"resident cold in test"_ and _"non resident"_ types are in test period, but _"resident cold"_ is not in its test period. So in the previous code, if _handHot_ met the _"resident cold"_ type of entry, the _runHandHot()_ method will fall into an infinite loop. As the author described in the paper, and also commented in the code, the most probable case is handHot met "non-resident" state of entry, so I think we were lucky that this code didn't falling into infinite loop in our previous tests. This change not impacts results because _terminateTestPeriod()_ method do nothing when the target node is not in test period.

ben-manes: oh great catch then, thank you 😄

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
chanyoung: Done.

chanyoung: Done.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
chanyoung: Done.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProPlusPolicy.java
chanyoung: Thank you for the comment! I didn't make change but that approach seems clearer and easier to use!

chanyoung:
> Please fix the analysis errors. In the new github actions CI, the script is .github/scripts/analyze.sh.

Fixed.

Issue #478, opened 2020-11-30
expireAfter(Expiry) does not reliably invoke removalListener


facboy:
`expireAfter(Expiry)` does not reliably invoke the `removalListener` when used with `expireAfterUpdate`.  test-case here: https://github.com/facboy/caffeine-test

as far as I can tell it happens when the `Expiry` only sets an expiry on `expireAfterUpdate`.  the `Node` ends up in the `TimerWheel` with its original `MAX_EXPIRY` and this is never updated so `getExpirationDelay()` always returns some value years in the future, and the expiry task is thus also scheduled for years in the future.

ben-manes:
Thanks for the test case! I think you're right, there is a mistake here when adding variable expiration that didn't fit well with the write tolerance handling. The logic here seems incorrect,

https://github.com/ben-manes/caffeine/blob/a54afaa9d00a78e335e5c2d7b62c5ce1b6ef82a3/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L2075-L2085

This causes the `expireAfter` to degrade into the `afterRead` case, which swallows the reordering and delays handling. A few minor changes to this if-block locally resolves the problem, but I need to review it more thoroughly before pushing and releasing for you. Hopefully will get to it tonight PST.

ben-manes:
Sorry, this is a very busy week for me. In the meantime you should be able to use `asMap.compute` as a workaround. I'll try to have a fix released by next Monday.

ben-manes:
I believe the right fix is to update the calculation to be aware of variable expiration.

https://github.com/ben-manes/caffeine/blob/a54afaa9d00a78e335e5c2d7b62c5ce1b6ef82a3/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L2054

should be replaced with

```java
exceedsTolerance = ((now - prior.getWriteTime()) > EXPIRE_WRITE_TOLERANCE)
    | (expiresVariable()
        && Math.abs(varTime - prior.getVariableTime()) > EXPIRE_WRITE_TOLERANCE);
```

This passes that failure of your test case, though it fails for what I think is a logic bug in the test itself.

[...]

PR #479, opened 2020-12-01
Ensure Expiry.expireAfterUpdate() triggers removalListener (#478)

facboy:


ben-manes:
Merged with additional unit tests

facboy:
i suppose one thing i forgot to highlight is that when i did the original PR i left the AND/OR shortcutting in:

https://github.com/ben-manes/caffeine/blob/e6bb02f4b0dd07a55a89a87334a6dc2815735507/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L2054

i guess your performance checks showed this didn't have an impact on performance?

ben-manes:
It shouldn't. Ideally the JIT should do dead code elimination and only one side of the condition would be evaluated, since either `expiresAfterWrite` or `expiresVariable` will always return false.

It also wouldn't be a perf difference regardless, as it was mostly a reminder to myself than an actual issue. So I hadn't disagreed, had done the perf testing on whether the tolerance was still a good idea, and used the cleaner syntax.

PR #496, opened 2021-02-08
Honor recordStats setting on get operation

johanbjork:
recordStats appears to be hardcoded to false on the get() method. Is it simply a bug? Or any other reason?

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
@@ -1866,7 +1866,7 @@ abstract class BoundedLocalCache<K, V> extends BLCHeader.DrainStatusRef<K, V>
 
   @Override
   public @Nullable V get(Object key) {
-    return getIfPresent(key, /* recordStats */ false);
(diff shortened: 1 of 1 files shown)


ben-manes:
This mirrors Guava, where `Map.get` does not record stats. Most map operations do not, except for computations. That avoids a few quirky cases, like `replace` or iterations, where it might not make sense. We tend to promote using a computation, ideally through a `CacheLoader`, so `Cache.get` is preferred and will record statistics.

johanbjork:
Hi,
Thanks for the enlightening response. The reason I started fiddling about with this is because [hazelcast-hibernate53](https://github.com/hazelcast/hazelcast-hibernate) just recently changed their local cache provider from a homebrewed ConcurrentHashMap based implementation, to Caffeine. Specifically [in this class](https://github.com/hazelcast/hazelcast-hibernate/blob/master/hazelcast-hibernate53/src/main/java/com/hazelcast/hibernate/local/LocalRegionCache.java).

I'm trying to extract cache statistics, but after your explanation, a better path is probably to refactor that code to operate on the Cache instead of its Map.

You can close this request then, thanks Ben!

johanbjork:
Aha, I can close it myself. Thanks.

PR #528, opened 2021-04-10
Added SYSTOR '17 trace files parsers

himelbrand:
I've added two parsers for the SNIA SYSTOR17 trace files to the simulator.
[http://iotta.snia.org/traces/4964](Trace files collection can be found here).
The 2 types of parsers are:

- Latnecy: using the response time as the miss penalty (multiplied by 1000 to get the value in milliseconds )
- Size: using the size field to create a weighted AccessEntry object.

I hope this contribution will help progress latency and size aware caching research.


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/systor/SystorTraceLatencyReader.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public Stream<AccessEvent> events() {
+
+    return lines()
+            .map(line -> line.split(","))
+            .filter(array -> array[2].equals("R") && array[1].length() > 0)
+            .map(array -> {
+              long key = new BigInteger(array[4]).longValue();
```
ben-manes, line 52: Just fyi that this truncates the number. For this workload I think it should be equivalent to being hashed, but I don't know if it causes skew compared to running it through a function like murmur3. You might want to try hashing to double check that that the hit rates are similar with either scheme.

```java
jshell> var b = BigInteger.valueOf(Long.MAX_VALUE).shiftLeft(2) // 4 x max
b ==> 36893488147419103228
jshell> var c = b.longValue()
c ==> -4
```

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/systor/SystorTraceLatencyReader.java
```diff
@@ -0,0 +1,58 @@
...
+              double responseTime = Double.parseDouble(array[1]);
+              responseTime *= 1000; //convert to milliseconds from seconds
```
ben-manes, line 54: I'm curious as to whether this has an impact since the penalty is a double without a unit. Would a policy behave any differently? I would naively assume it wouldn't change because the difference between two costs are the same.

ben-manes (approved):
Thanks for the PR. I'll merge it in after your replies to my questions, but those certainly are not blockers and just me being curious.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/systor/SystorTraceLatencyReader.java
himelbrand: You're righ about the possible truncation, I'm running a quick (as possible) analisys of the offsets field, checking the number of bits required to represent the largest offset, if it is less or equal than 64, then we are good with this implemantation.
If not, it might be better to use a hash.
I'll let you know as soon as Ihave the results of the analisys.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/systor/SystorTraceLatencyReader.java
himelbrand: You are correct regarding the fact that it still has the same proportion so it might not do any difference, however I came across approaches that uses a power of some degree on the cost/penalty, I've also experimented with a few approaches like those I mentioned, and for fractions it might cause underflow, so just for this reason I switched to milliseconds, for the SYSTOR17 traces it seems very reasonable since we still have pretty low values meaning the risk of overflow is quite low, and this also covers the risk for underflows.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/systor/SystorTraceLatencyReader.java
ben-manes: A simple approach is to use toByteArray and guava’s hasher. If it’s similar than no reason to change.

I suppose since this is not split into blocks (e.g. like ARC’s) the chance of a hit is much lower. A block trace would usually by unit sized and split by the block size, so there would be a chance for overlapping requests. An object style with weights would require the same offset was requested, which for disk traces is a bit low. I wonder if a block size should be used rather than a weight? The readme is vague but it seems odd to me when thinking through this.

ben-manes: Oh I was remembering the Umass reader which has 512 butte blocks
https://github.com/ben-manes/caffeine/blob/master/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/umass/storage/StorageTraceReader.java

ben-manes: It looks like modern disks increased to 4KiB block sizes a while back, so that was likely as this is a newer trace. If you don’t mind, I’ll probably switch to that and only keep a latency aware version. I think a weighted trace likely wouldn’t be a very realistic workload given how the disk caches actually work.

himelbrand: No problem, sounds very reasonable...
Do you want me to do it and then push the changes, or you will change it post merge?

ben-manes: I can do it post merge

himelbrand: and BTW I went over more than 50% of the offsets in all of the trace files and still no offset that is too big to fit in a long value (i.e. all of them so far are less than 64 bits)

ben-manes: Great, then let’s keep your truncation as seems fine and simple

himelbrand: I'm inclined to agree, since I assume that the traces are from a 64-bit system, thus all of the offsets should fit in 64 bits.

himelbrand: If after the analisys is over I find some problem with this, I could do a PR with a fix using a hash (I don't think it will be needed).
So on my side we can merge.

ben-manes: Thanks. It’s 1am PST so I’ll squash, merge, and do a pass of small changes tomorrow for both of your PRs. Thanks for the improvements.

PR #529, opened 2021-04-10
GdWheel policy + AccessEvent based admission

himelbrand:
I've Implemented GDWheel policy: [pdf](https://conglongli.github.io/paper/gdwheel-eurosys2015.pdf)
This policy is cost aware, where the cost can be either the weight or penalty (In the paper they use cost as the recomputation cost for an item, which is basicly the miss penalty).

In addition I've extended the Admittor so we now have EventsAdmittor, which uses the AccessEvent instead of just using the key, this is usful for size or latency based admission policies.


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/admission/Admittor.java
```diff
@@ -38,10 +40,41 @@
...
+
+  /**
+   * Returns if the candidate should be added to the cache and the page replacement policy's chosen
+   * victim should be removed.
+   *
+   * @param candidate the event of the newly added entry
+   * @param victim the event of the entry the policy recommends removing
+   * @return if the candidate should be added and the victim removed due to eviction
+   */
+  default boolean admit(AccessEvent candidate, AccessEvent victim) { return admit(candidate.key(),victim.key()); }
```
ben-manes, line 54: Please make these cases multiple lines (we casually follow [Google's style guide](https://google.github.io/styleguide/javaguide.html)). For languages like Kotlin / Scala this is common for one-liners, but isn't in Java.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/admission/Admittor.java
```diff
@@ -38,10 +40,41 @@
...
+
   /** Returns an admittor that admits every candidate. */
   static Admittor always() {
     return AlwaysAdmit.INSTANCE;
   }
+
+  /** An admission policy for the cache that uses the AccessEvent object as input instead of key,
+   * useful for admission policies that are based on size or penalties.
+   * **/
+  interface EventsAdmittor extends Admittor {
```
ben-manes, line 64: Maybe we should flip this around? The admittor was written prior to having an AccessEvent, where we introduced `KeyOnly` types for conversion. So we could make `KeyOnlyAdmittor` as a convenience for existing usages, while making the base interface event-aware.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
```diff
@@ -0,0 +1,235 @@
...
+            return config().getInt("gd-wheel.nw");
+        }
+
+        public int numberOfQueues() {
+            return config().getInt("gd-wheel.nq");
+        }
+    }
+
+    static class CostWheel {
+        List<PriorityQueue<AccessEvent>> wheel;
```
ben-manes, line 166: You might want to use `TreeSet` instead, which is implemented as a red-black tree. That has a O(lg n) operation for `remove(Object)`, whereas `PriorityQueue` is a heap with O(n) time.

> Implementation note: this implementation provides O(log(n)) time for the enqueuing and dequeuing methods
offer, poll, remove(), and add; linear time for the remove(Object) and contains(Object) methods; and constant time for the retrieval methods peek, element, and size.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/admission/Admittor.java
ben-manes: Actually, you can ignore this. I'll go back and fix up minor style nits (line Google's use of 2-space indent) after this is merged.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/admission/Admittor.java
ben-manes: If you're okay with this proposal, I can go back and migrate towards that. That way it doesn't dirty up your PR and be an isolated refactoring.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
```diff
@@ -0,0 +1,235 @@
...
+            return !queue.isEmpty();
+        }
+    }
+
+    static class EventComparator implements Comparator<AccessEvent> {
+        @Override
+        public int compare(AccessEvent e1, AccessEvent e2) {
+            return Double.compare(e1.missPenalty(),e2.missPenalty());
+        }
+    }
```
ben-manes, line 217: fyi, Java 8 introduced a one-liner for these cases by using lambdas, 
`Comparator.comparingDouble(AccessEvent:: missPenalty)`

ben-manes (approved):
Thanks for implementing this! I'll merge after the comments, and then go back to do some minor tidying up for consistency, etc.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
vlsi: PriorityQueue + HashMap all the way :)

ben-manes: In this case, `remove(Object)` is called on every cache hit to reorder the entry. Since that is an O(n) operation, the performance degrades as the cache size increases.

`PriorityQueue` is optimal when used a a queue, where specific entries are not commonly searched for. For example Java's `ScheduledExecutorService` uses a min-heap and marks the entry as dead when cancelled, rather then removing it. In that case the cheaper algorithmic cost is beneficial, especially since a scheduler rarely has millions of events. Here the remove + add to resort the entry on every hit could lead to higher execution times and make GDWheel appear slower than it might be in practice.

vlsi: I'm always forgetting that `PriorityQueue` misses `removeAt(int)` method, so adding `HashMap` does not solve the issue. BinaryHeap with node removal is needed.

Tombstones would create an ABA issue, so it might be easier to implement or borrow a binary heap if priorityqueue with removals is needed.

> Here the remove + add to resort the entry on every hit could lead to higher execution times and make GDWheel appear slower than it might be in practice

The paper suggests that Queue is a linked list:

>Since each queue is implemented by a doubly linked list, removing or inserting a given node requires only constant time

In other words, the queue needs:
* Fast append to `head`
* Fast removal from `tail`
* Fast removal from the middle

It is funny how `LinkedHashSet` almost works here. The sad part is `LinkedHashSet` does not expose `reverseIterator` even though it maintains double-linked list.

ben-manes: I'll have to reread the GDWheel paper, as it has been a few years. I believe Apache Commons' `ListOrderedMap` is that more flexible variant that would be simple to adapt here, based on your comment. In our [TimerWheel](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/TimerWheel.java) we use intrusive linked lists for expiration to maintain an O(1) priority queue. However that code is very optimized while this is simulator code for algorithm analysis. Ideally the simulator policies would be clean, readable reference implementations that are easy to adapt for others. They don't tackle messy topics like concurrency, though, and are strictly for analysis for exploring each algorithm's insights.

vlsi: `ListOrderedMap` uses `ArrayList` which implies `O(N)` removals.

What do you think of the following?
The below structure efficiently implements "add to head" (via `put(k, v)`), "remove from tail" (via `removeEldestEntry` trick) and "remove".

```java
public class DequeWithRemoval<K, V> extends LinkedHashMap<K, V> {
    private static final Object SENTINEL = new Object();
    private @Nullable V lastValue;
    private boolean removeLastRequested;

    public DequeWithRemoval(int initialCapacity, float loadFactor) {
        super(initialCapacity, loadFactor, /* accessOrder */ false);
    }

    public V removeLast() {
        if (isEmpty()) {
            throw new NoSuchElementException();
        }
        removeLastRequested = true;
        // put triggers removeEldestEntry which is how we retrieve the latest element
        //noinspection unchecked
        ((Map) this).put(SENTINEL, SENTINEL);
        V lastValue = this.lastValue;
        this.lastValue = null;
        // Remove sentinel
        values().iterator().remove();

        return lastValue;
    }

    @Override
    protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
        if (!removeLastRequested) {
            return false;
        }
        lastValue = eldest.getValue();
        return true;
    }
}
```

ben-manes: oh man, I didn't realize it was done so poorly in Apache. I suppose it's my fault for never using it an assuming it was on par if I ever found a need.

That's neat. I don't mind implementing intrusive linked lists. If I was to write implement this policy, I would include the prev/next references on the `Node` and handle it directly. This is done in many of the other policies for LRU ordering. The code is very simple and straightforward.

ben-manes: Oh, it was [LinkedMap](https://github.com/rkanter/collections-generic/blob/master/src/main/java/org/apache/commons/collections15/map/LinkedMap.java) that was like `LinkedHashMap`.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/admission/Admittor.java
himelbrand: Sounds good, I'm good with your proposal.
Thanks

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
himelbrand: I will change it to TreeSet, since according to the paper it should be O(log(n)) in the worst case and amortized is O(1), so O(n) for a single operation is not ideal for the implemantation.
Thanks for the info.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
himelbrand: Cool, I like this syntax better, I'll adopt it.

himelbrand:
I've just updated the few things you mentioned @ben-manes 
I pushed the changes now

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
vlsi: @himelbrand , could you please clarify where do you get `according to the paper it should be O(log(n))` from?

https://conglongli.github.io/paper/gdwheel-eurosys2015.pdf

> 3.2.2 GD-Wheel Time Complexity Analysis
>Since each
>queue is implemented by a doubly linked list, removing
>or inserting a given node requires only constant time. An
>insertion of a given object requires one list insertion. A
>reuse of a given object requires one list removal and one list
>insertion. Thus handling either an insertion or a reuse of an
>object takes only O(1) time in the worst case

I do not see a priority queue here. However, I must admit it is the first time I see the paper, so I can be wrong.

himelbrand: @vlsi, the are 2 published versions of this paper, but I think this appears in both of them.
specifically regarding the priority queue, it is definitely mentioned in the pseudo code of the algorithm and a few extra locations.

regarding the time complexity, if you keep going a bit further in 3.2.2 you'll see that O(1) is amortized. I might have been confused, as I now see that they mentioned O(n), but I might remember the log(n) from the other version when compared to the gree dual approach.

vlsi: PriorityQueue is mentioned in `Algorithm 1 The GreedyDual Algorithm under Cao et al.’s implementation`
However, point of adding `GD-Wheel` policy is to replace priority queues with multiple linked lists.

O(n) is related to `migration` which indeed scans the queue.

>paper: As a result, **objects with different priorities** are stored in **different queues**

So I'm inlined that the queues should be implemented as linked lists so the implementation matches the paper.

himelbrand: When I implemented it I did it according to [pdf](https://www.cs.cmu.edu/~conglonl/gdwheel-ladis2013.pdf).
which is the Algorithm 1 The GreedyDual Algorithm under Cao et al.’s implementation.
the paper I gave the link and that you cited, was not the one I was reading when I implemented the algorithm.
This is the reason for the confusion.

It could be nice to implement the other version too, or to add the possibility to switch between the 2 variants in the future.

ben-manes: Let me try to spend some time reading the two papers and seeing if I can optimize it while matching the hit rates. I was planning on refactoring the code a bit to pull apart the long `record` method, e.g. to match the other policies which use `onHit`, `onMiss`, `evict`, etc. I'd also like to extract paper comments to help guide the reader, e.g. as done in `GdfsPolicy`. This will all take a bit longer, but the code is a good baseline to refactor on top of.

ben-manes:
@himelbrand I quick sanity test and I think the behavior is broken somehow. I was just doing minor formatting changes prior to reading the paper(s), and added a validation check which failed. For the default lirs trace, multi1, it reports a map size of `2606` whereas the maximum is `512`. I reproduced this with your commit and w/o my formatting changes.

```java
@Override
public void finished() {
  checkState(data.size() <= maximumSize, "%s > %s", data.size(), maximumSize);
}
```

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/GDWheelPolicy.java
```diff
@@ -0,0 +1,227 @@
...
+                    break;
+                }
+            }
+            return current_index;
+        }
+
+        public AccessEvent evict(int i) {
+            TreeSet<AccessEvent> queue = wheel.get(i);
+            return Objects.requireNonNull(queue.first());
+        }
```
ben-manes, line 193: This is the bug that causes the cache to not evict correctly. The method`first()` peeks into the queue, whereas you want `pollFirst()` to remove it.

ben-manes: This fix only works if always allowing admission, as if `admit = false` then it seems to break...

ben-manes:
I rewrote the implementation onto O(1) intrusive lists as described by the paper, where I used a sentinel node for simplicity. The algorithm's steps are annotated from the pseudo code to help future readers follow along. For an unknown reason, I did observe a large hit rate improvement on `mutli1.trace` when comparing against the original version, and it now more closely matches GdfsPolicy's hit rate. I removed the usage of an admission policy since I don't think you were actually using it and that caused weird failures in your version.

This was a bit more invasive of a refactor than I had expected, so I apologize for that. I think it more closely matches the paper and is clearer to follow. It helped a lot to have yours and the paper as a reference.

vlsi:
Just in case, it turns out `LinkedHashMap` is good enough (see the updated https://github.com/ben-manes/caffeine/pull/529#discussion_r611101853)
`removeLast` is a mere `iterator().next().remove()`. `removeEldestEntry` is not really needed.

PR #544, opened 2021-04-27
Update GDWheelPolicy.java

himelbrand:
A fix for null pointer in migrate.
changed the condition of the if statment as to not cause a null pointer exception by trying to access wheel at index that is out of bounds.


ben-manes:
Thanks!

PR #550, opened 2021-05-16
Add simplified ClockPro policy

chanyoung:
It is a simple implementation of ClockPro, but using more memory (8bytes per metadata).


ben-manes (approved):
We follow [Google's style guide](https://github.com/google/styleguide), though don't strictly enforce it. A minor nit though is 100c line lengths, whereas your comments follow a 120c length. For consistency can you please reformat those doc strings?

Do you want your new policy listed in `reference.conf`? If not, it may not be discoverable.

ben-manes:
Please run the static analyzers. Looks like you can make anode a static nested class.

chanyoung:
> We follow [Google's style guide](https://github.com/google/styleguide), though don't strictly enforce it. A minor nit though is 100c line lengths, whereas your comments follow a 120c length. For consistency can you please reformat those doc strings?

Done. 

> Do you want your new policy listed in `reference.conf`? If not, it may not be discoverable.

Added the policy into the `reference.conf`.

> Please run the static analyzers. Looks like you can make anode a static nested class.

Sorry, I forgot to run static analyzers. Done.

PR #553, opened 2021-05-20
fix #552: NodeFactory and LocalCacheFactory avoid reflection

### LINKED ISSUE #552: GraalVM native-image and LocalCacheFactory/NodeFactory
Hi, I'm sorry if this has been discussed elsewhere, but I've recently run into some foot-guns using native-images with caffeine caches.

I've been able to work around the reflection issues by adding reflection metadata for constructors of all implementations of `LocalCacheFactory` and `NodeFactory`, however that adds a lot of data to my image (5-10 mb). I see a similar issue on quarkus here, where only a subset of configurations are supported:
https://github.com/quarkusio/quarkus/blob/2cbd177416f92b9f45abb34b6f01acb8cb0f4eb5/extensions/caffeine/deployment/src/main/java/io/quarkus/caffeine/deployment/CaffeineProcessor.java#L19-L33

Would it be reasonable to generate references to constructors inline instead of using `Class.forName` and varhandles? The factory method would contain more bytecode to enumerate constructors, but I wouldn't expect it to be terribly large, equivalent to a few hundred lines of java code. I can put together a proof of concept if that's helpful.

Thanks!

carterkozak:


ben-manes:
Thanks. I'm closing for now since it is unclear what the right steps are. Let's revisit once there is more clarity.

PR #561, opened 2021-06-01
CAMP - greedy dual size based policy implementation 

himelbrand:
I've Implemented CAMP policy: [pdf](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.466.393&rep=rep1&type=pdf) for the simulator.

Since the simulator has 2 greedy-dual based, I thought if adding another that I think is pretty interesting, and I also saw that @ben-manes mentioned this policy on a discussion on time/penalty/benefit/cost aware caching.

I tried using the Google java style XML this time, I hope it worked properly.


simulator/src/main/resources/reference.conf
```diff
@@ -452,6 +452,11 @@ caffeine.simulator {
     queues = 256
   }
 
+  camp {
+  # Parameter precision p
+  p = 5
```
ben-manes, line 457: Please call the parameter `precision` and the Java code to use that variable name instead of the non-descriptive `p`.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/CampPolicy.java
```diff
@@ -0,0 +1,280 @@
...
+import com.google.common.base.MoreObjects;
+import com.typesafe.config.Config;
+
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+
+/**
+ * CAMP algorithm. Greedy Dual Wheel (GD-Wheel) algorithm.
```
ben-manes, line 39: I think remove the second sentence?

ben-manes (approved):
Thanks! I'll re-read the paper tonight to better follow along your logic, but at a glance it looks pretty clean. When simple enough to do, please annotate the class with notes from the paper to help others follow along.

A recent [paper](https://www.ymsir.com/papers/chopt-sigmetrics.pdf) ([video](https://www.youtube.com/watch?v=4pOwjwITZxc]) that might be of interest brute forces the optimal value for cost-aware policies using a min/max flow.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/CampPolicy.java
himelbrand: Yes, I'll removed it, made a commit + push now

simulator/src/main/resources/reference.conf
himelbrand: Changed it to `precision` in both reference and in the code itself

himelbrand:
I added some comments through the code (not to many) I hope it helps to make it clearer + changing `p` to `precision`

PR #606, opened 2021-10-10
add design.png for caffeine

huzais520:


huzais520:
Hi Ben, I have changed the design based on your suggestions.

1. Eden → Window
2. degrade → downgrade
3. Hill climber → added it to the data size description
4. Source format → I redraw the design with Microsoft Visio. Shall I commit it as well? 

As for the TimerWheel, I find it a little bit hard to add it because there is no more space in the png. Sorry about that.

Issue #7, opened 2015-02-27
Bulk refresh


ben-manes:
This [Guava issue](https://github.com/google/guava/issues/1975) identified an expected optimization not being implemented. A `getAll` where some of the entries should be refreshed due to `refreshAfterWrite` schedules each key as an independent asynchronous operation. Due to the provided `CacheLoader` supporting bulk loads, it is reasonable to expect that the refresh is performed as a single batch operation.

This optimization may be invasive and deals with complex interactions for both the synchronous and asynchronous cache implementations. Or, it could be as simple as using `getAllPresent` and enhancing it to support batch read post-processing.

ben-manes:
The API for this improvement is not very clear. Even if a batch refresh may predominantly delegate to `loadAll(keys)`, the method should try to be symmetric with the `reload(key, @Nonnull value)`. What would the batch version of that be?

A batch refresh using `refreshAfterWrite` and triggered by `getAll` would maps directly to `reloadAll(Map<K, V> entries)`. However, `LoadingCache.refresh(key)` may will `reload` if the entry is present and `load` if it is not. How should `refreshAll(keys)` behave when some of the entries are present and some are not? Should it split the operation into two calls, `reloadAll(present)` and `loadAll(missing)`, or should it delegate the decision to the user, e.g. `reloadAll(Map<K, V> entries, Iterable<K> remainingKeys)`? The latter seems wrong from an API consistency perspective, but correct from a performance one.

I'm favoring the `reloadAll(Map)` approach because it is more consistent and we can optimize the majority of the use-cases. Here's my rational.
 * A partially present batch is not the common case (explicit refresh). Most usages will be triggered by using `refreshAfterWrite` and a batch `getAll`.
 * Most users do not use the old values when computing the new one, relying on the default implementation (delegate to `load`).
 * Multiple operation are only required if the user overrides the default implementation (delegate to `loadAll`). This can be detected using reflection (at cache construction), allowing us to perform a single call instead of two since that will have the same user visible behavior.
 * A refresh is performed asynchronously, so 2 is a better worse case than N.
 * Like `getAll`, we cannot ensure that parallel batch operations do not overlap and may load the same entries. In the common case of refreshAfterWrite this can be optimized to not occur, though.

On reflection, I like `CacheLoader.reloadAll(Map)` as providing the best power-to-weight ratio.

@abatkin @cjohnson2

abatkin:
[...]

PR #620, opened 2021-11-01
Some cleanup

sheepdreamofandroids:
in response to https://github.com/ben-manes/caffeine/issues/7#issuecomment-956281417

### DIFF
--- a/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
+++ b/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
@@ -76,8 +76,10 @@ public class CoalescingBulkloader<Key, Value> implements AsyncCacheLoader<Key, V
      *                    Extra values are ignored. Missing values lead to a {@link java.util.NoSuchElementException}
      *                    for the corresponding future.
      */
-    public static <Key, Value> CoalescingBulkloader<Key, Value> byOrder(int maxLoadSize, long maxDelay,
-                                                                        final Function<Stream<Key>, CompletableFuture<Stream<Value>>> load) {
+    public static <Key, Value> CoalescingBulkloader<Key, Value> byOrder(
+            int maxLoadSize,
+            long maxDelay,
+            final Function<Stream<Key>, CompletableFuture<Stream<Value>>> load) {
         return new CoalescingBulkloader<Key, Value>(maxLoadSize, maxDelay, toLoad -> {
             final Stream<Key> keys = toLoad.stream().map(wk -> wk.key);
             load.apply(keys).thenAccept(values -> {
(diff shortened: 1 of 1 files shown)


examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
```diff
@@ -124,9 +129,9 @@
      *                       for the corresponding future.
      */
     public static <Key, Value, Intermediate> CoalescingBulkloader<Key, Value> byExtraction(int maxLoadSize, long maxDelay,
-            final Function<Intermediate, Key> keyExtractor,
-            final Function<Intermediate, Value> valueExtractor,
-            final Function<Stream<Key>, CompletableFuture<Stream<Intermediate>>> load) {
+                                                                                           final Function<Intermediate, Key> keyExtractor,
+                                                                                           final Function<Intermediate, Value> valueExtractor,
+                                                                                           final Function<Stream<Key>, CompletableFuture<Stream<Intermediate>>> load) {
```
ben-manes, line 134: looks like a formatting mistake?

ben-manes: oh nm, I see, to line up. just fyi we're loosely using Google's java style if that helps, but I am not particular on these examples.

examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
```diff
@@ -48,18 +49,22 @@
...
     private ScheduledFuture<?> schedule;
     // Queue.size() is expensive, so here we keep track of the queue size separately
-    private AtomicInteger size = new AtomicInteger(0);
+    private final AtomicInteger size = new AtomicInteger(0);
 
-    private final class WaitingKey {
-        Key key;
-        CompletableFuture<Value> future;
-        long waitingSince;
+    class WaitingKey {
```
ben-manes, line 60: I think this can be a static final class? It doesn't seem to need to retain the hidden reference to the parent.

examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
sheepdreamofandroids: Not if you make the window big enough...
The formatter lines up everything after the opening bracket which is a bit far right in this case.

ben-manes (approved):
let me know when ready to merge. only change is the inner vs nested class is possible.

examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
sheepdreamofandroids: I forced the parameters to separate lines

examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
sheepdreamofandroids: Tried that. Then it doesn't get the Key type parameter.
Let me try giving it it's own parameter.

ben-manes: oh yes, you would need to define as `WaitingKey<K>` and use the diamond operator when newing it

sheepdreamofandroids: Yes, with it's own type parameters it can be static

ben-manes: nice. ready for merge?

sheepdreamofandroids:
I need to fix the email address on the commits for the CLA assistant.

sheepdreamofandroids:
I'm going to recreate this PR using the correct email address if necessary. But I'll do that tomorrow, it's getting late now.

ben-manes:
no worries, thank you 🙂

PR #621, opened 2021-11-02
same fixes as yesterday

sheepdreamofandroids:
for https://github.com/ben-manes/caffeine/issues/7#issuecomment-956281417

made more fields final
applied google formatter
gave the tests more slack to also succeed consistently on Windows 10

### DIFF
--- a/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
+++ b/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
@@ -15,10 +15,8 @@
  */
 package com.github.benmanes.caffeine.examples.coalescing.bulkloader;
 
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static java.util.stream.Collectors.toMap;
-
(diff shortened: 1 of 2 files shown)


examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
```diff
@@ -37,155 +35,185 @@
...
+            load.apply(keys)
+                .thenApply(
+                    intermediates -> intermediates.collect(toMap(keyExtractor, valueExtractor))));
+  }
+
+  private CoalescingBulkloader(
+      int maxLoadSize, long maxDelay, Consumer<Collection<WaitingKey<Key, Value>>> bulkLoader) {
+    this.bulkLoader = bulkLoader;
+    assert maxLoadSize > 0;
+    assert maxDelay > 0;
```
ben-manes, line 166: purely fun fyi, but `assert` lost favor in Java's core libs where it used to be popular. This is because it increased the bytecode size compared to a regular check, which increased the chance for a method to exceed the inlining cliff (35 bytes). The keyword used to be used liberally in j.u.c., if I recall correctly, when originally written to favor readability. It was stripped out a few releases later when maximizing for performance instead.

PR #690, opened 2022-03-25
Refactored StipedBuffer.java

shivashankar0601:
Hi,

I Performed refactoring on **expandOrRetry** method where the method seems to be bloating and the cyclomatic complexity was high. I have moved a part of the code to a newly created method so that the complexity is reduced and the size of the method is reduced by increasing the readability and understandability.

I have renamed a few variables whose name was not meaning for the context. I have renamed the variable "z" to threadID and variable "h" to modifiedThreadID which seem meaningful to the context.

Note:
performed these refactoring out of curiosity and academic purpose.

Thanks,
Shiva.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/StripedBuffer.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/StripedBuffer.java
@@ -31,7 +31,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * A base class providing the mechanics for supporting dynamic striping of bounded buffers. This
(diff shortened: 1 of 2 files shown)


ben-manes:
I appreciate this and I understand the reasons. It is something that I would do too if this was originally my code. However, it originates from Doug Lea's Striped64 class which is part of the JDK for use by `LongAdder`. That includes some of the adaptions, such as the hashing of the thread id, which came from private communications. In this case I would prefer to keep it in the style of the original author's. That makes it easier when working with them on bugs and sync'ing their improvements. For that reason I try to make minimal modifications, because while your changes help readability it is more jarring for those authors.

PR #691, opened 2022-03-25
refactored StripedBuffer.java to reduce the methods cyclomatic complexity and renamed variables to give meaningful name to them

shivashankar0601:
Hi,

I refactored **StripedBuffer.java** where the method **expandOrRetry** was bloating and was having high cyclomatic complexity. I refactored it by creating two new private methods **isCreated** and **isInit** that support the method. This reduced the size of the method by increasing the readability and understandability of the code.

I also refactored a few variables whose name was not appropriate for the context. For example, the variable **"z"** was not giving any contextual meaning hence I refactored it to **"threadID"**. The other variable **"h"** was also not giving the contextual meaning, hence refactored it to **"modifiedThreadID"**.

**Note:**
I have done this out of curiosity to learn how to contribute to open-source projects and as a part of my academic work. Hope this is useful.

Thanks,
Shiva.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/StripedBuffer.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/StripedBuffer.java
@@ -31,7 +31,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * A base class providing the mechanics for supporting dynamic striping of bounded buffers. This
- * implementation is an adaption of the numeric 64-bit {@link java.util.concurrent.atomic.Striped64}
+ * implementation is an adaption of the numeric 64-bit { @link java.util.concurrent.atomic.Striped64 }
  * class, which is used by atomic counters. The approach was modified to lazily grow an array of
  * buffers in order to minimize memory usage for caches that are not heavily contended on.
  *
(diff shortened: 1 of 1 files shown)


ben-manes:
I think you missed my [reply](https://github.com/ben-manes/caffeine/issues/690#issuecomment-1078583892_) in your prior PR,

> I appreciate this and I understand the reasons. It is something that I would do too if this was originally my code. However, it originates from Doug Lea's Striped64 class which is part of the JDK for use by `LongAdder`. That includes some of the adaptions, such as the hashing of the thread id, which came from private communications. In this case I would prefer to keep it in the style of the original author's. That makes it easier when working with them on bugs and sync'ing their improvements. For that reason I try to make minimal modifications, because while your changes help readability it is more jarring for those authors.

shivashankar0601:
Hi,

There seem to be a problem with one of the commits. I already checked my code and already read the log. Non of the code I have written is causing this issue. Kindly let me know if I can do anything for this.

Thanks,
Shiva.

shivashankar0601:
Hello,

Thanks a lot for your reply, I understand the situation. I tried my best to refactor it. As I mentioned earlier, I am doing this as a part of my academics, If the code I push is not relevant and is not needed or is making any conflicts with the interest of the project, kindly ignore it.

Thanks,
Shiva.

ben-manes:
I'll close because other factors favor leaving it as is. Thank you, though!

---

The code scanner, LGTM, is trying to force the build to be run by `java-8-openjdk-amd64` but the project is Java 11-based. Gradle is configured to auto-provision the JVM if needed, so I'm not sure how they overrode that behavior. There is probably some configuration somewhere, but I think Github's CodeQL replaces this so we might remove the integration.

PR #697, opened 2022-04-14
Fix a thread-safe issue of schedule.

luozhenyu:
# Description

The variable "schedule" is accessed by multiple threads. The usage of schedule in synchronized block has memory visibility to other threads, but the usage out of synchronized does not. Therefore, the variable "schedule" should be decorated with volatile or AtomicReference.

# Not thread-safe code

The variable "schedule" is written is synchronized block but read out of synchronized block. The visibility is not guaranteed.

https://github.com/ben-manes/caffeine/blob/d9f132cefe9a0833686354bd9986832e4f7d6beb/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java#L176-L180

# Performance consideration

Original code: synchronized block.

https://github.com/ben-manes/caffeine/blob/d9f132cefe9a0833686354bd9986832e4f7d6beb/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java#L185-L188

Alternative: CAS has better performance than synchronized block.

```java
  private void startWaiting() {
    ScheduledFuture<?> oldSchedule = schedule.getAndSet(timer.schedule(this::doLoad, maxDelay, MILLISECONDS));
    if (oldSchedule != null) {
      oldSchedule.cancel(false);
    }
  }
```

### DIFF
--- a/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
+++ b/examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader/CoalescingBulkloader.java
@@ -31,6 +31,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


ben-manes:
Thanks for the detailed explanation and fix! I’m not very familiar with this example code, @sheepdreamofandroids can you take a look?

sheepdreamofandroids:
At any time there is at most one schedule that is not done, therefore startWaiting() might be called too often but never too little. That is not clearly documented though and can easily break when modifying.

Your code is more explicit.

PR #698, opened 2022-04-14
Reorder the operation of reducing size before that of bulk loading

luozhenyu:
# Description

Bulk loading is expected to be a time-consuming operation. During bulk loading, other threads may invoke the method "doLoad" which is unnecessary, because the queue size is still larger than maxLoadSize. These threads will invoke "doLoad" immediately after finishing pervious "doLoad" but load nothing.

So moving the reduce-queue-size operation before the bulk-load operation will not cause the problem above.


sheepdreamofandroids:
Good catch!

PR #724, opened 2022-06-04
Change Description of feature.

sinsuren:
Hi @ben-manes 

Above documentation creates confusion, even I assumed that by default `key/value` are being wrapped with `WeakReference`.

The word `automatically` means, without any user intervention.
Screenshot 2022-06-04 at 5 40 22 PM

Due to above wording in documentation, sometimes it might lead to same confusion again.


sinsuren:
https://github.com/ben-manes/caffeine/pull/724

ben-manes:
I added "optional" into leading description, which should clarify that none of those features must be applied

PR #747, opened 2022-07-30
Use initial capacity for linked hash set in getAllPresent()

bbenshalom:
I noticed that under intensive load, there is a performance penalty from the constant resizing LinkedHashSet in getAllPresent(). Since this was addressed in the master branch, I saw fit to use the same code from there and bring this improvement to v2.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java
@@ -23,6 +23,7 @@ import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.time.Duration;
 import java.util.ConcurrentModificationException;
+import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


ben-manes:
There are a couple of other places in the [original commit](https://github.com/ben-manes/caffeine/commit/2649ac940e89505561808cf4c755e1aa1f8172f3), such as the `Cache.getAllPresent`. If we were to include this optimization, then we'd want to apply it in all of the usages.

ben-manes:
Also in an issue I recently stated,

> I am hoping to softly deprecate 2.x as jdk 8 is it has reached end of life (Mar 2022) and will stop receiving critical security fixes in 2025. That means there is a higher burden on companies who chose to stay on Java 8 and an expectation of diminishing industry support.

In this case, a company using Java 8 is ignoring many much more significant performance improvements available in later JDKs. If that's acceptable, then minor changes like this is hard to justify. Otherwise it is fixed in the latest version and you can upgrade. Therefore I am hesitant to invest into improving v2 except where users are impacted by bugs. Does that sound reasonable?

bbenshalom:
@ben-manes sounds completely reasonable. I'll close this, thanks.

Issue #750, opened 2022-08-03
Cannot consume in Android build that uses R8


nrobi144:
Using https://github.com/aPureBase/KGraphQL (`:0.17.14-fork-7`), that has a dependency on `com.github.ben-manes.caffeine:caffeine:3.0.2` in an Android project results in 

```
com.android.tools.r8.internal.ez0: MethodHandle.invoke and MethodHandle.invokeExact are only supported starting with Android O (--min-api 26)
```

nrobi144:
This error can be fixed by upgrading Android `minSdkVersion to 26`.

However
```
com.android.tools.r8.internal.ez0: Call to polymorphic signature of VarHandle are only supported starting with Android P (--min-api 28)
``` 

occurs on `minSdk < 28`

ben-manes:
This project does not support Android. The design goals for a server-side cache is very different from mobile, e.g. concurrency, dependency size, and memory assumptions. After we wrote Guava's cache, the Android team wrote LruCache as its counterpart because Google recognized that the ecosystems were different. For example, [Store](https://github.com/dropbox/Store) is an Android cache that is tailored for those users. This was true pre-Android too, with Java ME being only a subset of Java SE to avoid performance pitfalls.

A library may be lucky to work in both ecosystems without any changes, but this is simply good fortune. As the two continue to diverge more then library authors who want to support both will need to offer that via adapters.

ben-manes:
fwiw, the twitch4j project ran into this and created a common api, [Xanthic/cache-api](https://github.com/Xanthic/cache-api).

PR #751, opened 2022-08-03
Remove FactoryBenchmark.java to avoid Android desugaring issue

nrobi144:
Fixes `com.android.tools.r8.internal.ez0: MethodHandle.invoke and MethodHandle.invokeExact are only supported starting with Android O (--min-api 26)`

To fix https://github.com/ben-manes/caffeine/issues/750 `ManyToOneSpacedBuffer.java` might need a different implementation without `VarHandle`s


ben-manes:
Closing per other comments. The 2.x release does not use method handles as it targets Java 8.

PR #752, opened 2022-08-03
Replace VarHandle with AtomicReferenceArray

nrobi144:
Adds a possible fix for https://github.com/ben-manes/caffeine/issues/750#issuecomment-1203670570


ben-manes:
The primary reason for VarHandles is for the code generated `Node` classes, which is the cache entry. This holds the value, expiration timestamp, lru prev/next pointers, etc. Otherwise we must either waste memory (`AtomicReference`), use reflection (`AtomicReferenceFieldUpdater`), or use unsupported features (`sun.misc.Unsafe`). We chose the unsafe option in v2.x (Java 8) and VarHandles in 3.x. Both are supported.

PR #814, opened 2022-11-08
Filter invalid event in object store trace

guykhazma:
The object store trace contains some range reads events with invalid ending offset leading to events with negative weight.
For example, the following events in `IBMObjectStoreTrace001Part0`:
```
192199064 REST.GET.OBJECT 6c44b4d75e985008 0 0 -1
192199064 REST.GET.OBJECT fb7caa17fb475a5c 0 0 -1
192200755 REST.GET.OBJECT 12c21c4f26736f15 0 0 -1
192200755 REST.GET.OBJECT a8a514e89f0ff7aa 0 0 -1
```

This PR filters these events as part of the reader to avoid pre-processing the input file


ben-manes:
nice find, thanks!

PR #816, opened 2022-11-09
fix minor issue with array length condition in object storage traces

guykhazma:
This PR fixes a minor issue with the condition checking for GET requests without range reads.

In addition, since the reader filters only GET requests, I am not sure that the condition is needed at all, I haven't seen any GET request in the traces without range reads so it seems that this condition was never met anyway.
Maybe @ohadeytan knows if there are such cases.


ohadeytan:
Good catch. I believe they weren't such cases, so the condition is indeed redundant.

ben-manes:
nice, can you delete / rewrite the code @guykhazma to what you think is a clean and correct parser? I'll merge once you give me the okay.

guykhazma:
@ben-manes done, I have removed the unnecessary code, thanks.

PR #857, opened 2023-01-10
Include the junit platform via a test implementation

ThePumpingLemma:
This prevents junit-bom from becoming a dependency of caffeine's published artifacts

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,6 +89,7 @@ subprojects {
     testImplementation testLibraries.awaitility
     testImplementation testLibraries.osgiCompile
     testImplementation platforms.collect { platform(it) }
+    testImplementation testPlatforms.collect { platform(it) }
 
     testRuntimeOnly testLibraries.osgiRuntime
     testRuntimeOnly testLibraries.junitEngines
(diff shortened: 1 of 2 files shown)


ben-manes:
This looks like a good change (especially the typo fix!). I don't believe Gradle publishes the test scoped configurations. I realized my goof through in the accidental leak which should be resolved in https://github.com/ben-manes/caffeine/commit/396a9405f52d8f7bd3baa162fb6f1a9f5a239c91 and, then more thoroughly, in https://github.com/ben-manes/caffeine/commit/7321a56a0fe92308c5b20c2a789ab77907ca480e.

build.gradle
```diff
@@ -88,7 +88,7 @@ subprojects {
     testImplementation testLibraries.hamcrest
     testImplementation testLibraries.awaitility
     testImplementation testLibraries.osgiCompile
-    testImplementation platforms.collect { platform(it) }
+    testImplementation testPlatforms.collect { platform(it) }
```
ben-manes, line 91: I think this would be both platforms + testPlatforms? Since now the constraints are not applied universally like before, I am unsure if the separation is as important. But I do think it would be additive rather than exclusive, at least.

ThePumpingLemma: Fixed! In my projects, I tend to enable consistent version resolution, but I shouldn't assume that here.

PR #868, opened 2023-02-13
Add `@CanIgnoreReturnValue` annotations

Stephan202:
Suggested commit message:
```
Add `@CanIgnoreReturnValue` annotations

This cancels out the package-level `@CheckReturnValue` annotations
introduced in ff1238567ede4001270ac6ed4682989b6b508c7e for selected
methods.

While there, add `@CheckReturnValue` to all remaining packages.
```

Reason for filing this PR, is that when upgrading to Caffeine 3.1.3 our build fails with
```
[WARNING] /path/to/CacheFactory.java:[365,59] [CheckReturnValue] The result of `recordStats(...)` must be used
  If you really don't want to use the result, then assign it to a variable: `var unused = ...`.

  If callers of `recordStats(...)` shouldn't be required to use its result, then annotate it with `@CanIgnoreReturnValue`.

    (see https://errorprone.info/bugpattern/CheckReturnValue
 Did you mean 'statsCounter.ifPresent(counter -> caffeine = caffeine.recordStats(() -> counter));'?
```
The above error is triggered by this code:
```java
statsCounter.ifPresent(counter -> caffeine.recordStats(() -> counter));
```

(Our build has nearly all Error Prone checks enabled.)

### DIFF
--- a/caffeine/build.gradle
+++ b/caffeine/build.gradle
@@ -122,7 +122,6 @@ tasks.named('compileJava').configure {
   dependsOn generateLocalCaches, generateNodes
   finalizedBy compileCodeGenJava
   options.errorprone.disable('CheckReturnValue')
-  options.errorprone.error('CanIgnoreReturnValueSuggester')
 }
 tasks.named('compileTestJava').configure {
   dependsOn jar, compileCodeGenJava
--- a/caffeine/build.gradle
+++ b/caffeine/build.gradle
@@ -122,6 +122,7 @@ tasks.named('compileJava').configure {
   dependsOn generateLocalCaches, generateNodes
   finalizedBy compileCodeGenJava
   options.errorprone.disable('CheckReturnValue')
+  options.errorprone.error('CanIgnoreReturnValueSuggester')
 }
 tasks.named('compileTestJava').configure {
   dependsOn jar, compileCodeGenJava
--- a/guava/src/main/java/com/github/benmanes/caffeine/guava/CaffeinatedGuavaLoadingCache.java
+++ b/guava/src/main/java/com/github/benmanes/caffeine/guava/CaffeinatedGuavaLoadingCache.java
@@ -123,7 +123,7 @@ final class CaffeinatedGuavaLoadingCache<K, V>
   }
 
   @Override
-  @SuppressWarnings({"CheckReturnValue", "FutureReturnValueIgnored"})
+  @SuppressWarnings("FutureReturnValueIgnored")
   public void refresh(K key) {
     cache.refresh(key);
   }
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
@@ -73,6 +73,7 @@ import com.github.benmanes.caffeine.jcache.management.JCacheStatisticsMXBean;
 import com.github.benmanes.caffeine.jcache.management.JmxRegistration;
 import com.github.benmanes.caffeine.jcache.management.JmxRegistration.MBeanType;
 import com.github.benmanes.caffeine.jcache.processor.EntryProcessorEntry;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 
 /**
  * An implementation of JSR-107 {@link Cache} backed by a Caffeine cache.
@@ -279,7 +280,6 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   /** Performs the bulk load where the existing entries are replaced. */
-  @SuppressWarnings("CheckReturnValue")
   private void loadAllAndReplaceExisting(Set<? extends K> keys) {
     Map<K, V> loaded = cacheLoader.orElseThrow().loadAll(keys);
     for (var entry : loaded.entrySet()) {
@@ -288,7 +288,6 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   /** Performs the bulk load where the existing entries are retained. */
-  @SuppressWarnings("CheckReturnValue")
   private void loadAllAndKeepExisting(Set<? extends K> keys) {
     List<K> keysToLoad = keys.stream()
         .filter(key -> !cache.asMap().containsKey(key))
@@ -351,6 +350,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
    * @param publishToWriter if the writer should be notified
    * @return the old value
    */
+  @CanIgnoreReturnValue
   protected PutResult<V> putNoCopyOrAwait(K key, V value, boolean publishToWriter) {
     requireNonNull(key);
     requireNonNull(value);
@@ -474,6 +474,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
    * @param publishToWriter if the writer should be notified
    * @return if the mapping was successful
    */
+  @CanIgnoreReturnValue
   private boolean putIfAbsentNoAwait(K key, V value, boolean publishToWriter) {
     boolean[] absent = { false };
     cache.asMap().compute(copyOf(key), (k, expirable) -> {
@@ -552,6 +553,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   @Override
+  @CanIgnoreReturnValue
   public boolean remove(K key, V oldValue) {
     requireNotClosed();
     requireNonNull(key);
@@ -1238,7 +1240,6 @@ public class CacheProxy<K, V> implements Cache<K, V> {
     }
 
     @Override
-    @SuppressWarnings("CheckReturnValue")
     public void remove() {
       if (current == null) {
         throw new IllegalStateException();
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
@@ -444,6 +444,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   @Override
+  @CanIgnoreReturnValue
   public boolean putIfAbsent(K key, V value) {
     requireNotClosed();
     requireNonNull(value);
@@ -508,6 +509,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   @Override
+  @CanIgnoreReturnValue
   public boolean remove(K key) {
     requireNotClosed();
     requireNonNull(key);
@@ -621,6 +623,7 @@ public class CacheProxy<K, V> implements Cache<K, V> {
   }
 
   @Override
+  @CanIgnoreReturnValue
   public boolean replace(K key, V oldValue, V newValue) {
     requireNotClosed();
     requireNonNull(oldValue);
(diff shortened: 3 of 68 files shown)


caffeine/build.gradle
```diff
@@ -122,6 +122,7 @@ tasks.named('compileJava').configure {
   dependsOn generateLocalCaches, generateNodes
   finalizedBy compileCodeGenJava
   options.errorprone.disable('CheckReturnValue')
+  options.errorprone.error('CanIgnoreReturnValueSuggester')
```
ben-manes, line 125: This might go into codeQuality.gradle?
https://github.com/ben-manes/caffeine/blob/d084f2d3f1bb210dcba5b3f054039680178ef707/gradle/codeQuality.gradle#L154-L155

ben-manes:
@kluever please review. I guess that I missed that this was added to all the build methods. I just want to confirm this is everyone's intent.

caffeine/build.gradle
Stephan202: Ah, I missed that. Lemme try!

Stephan202: This way it also flags some test code. That's fine; will add the annotations there too.

ben-manes:
I suppose we can also remove some of the suppressions that are now irrelevant. I think that would be [CaffeinePolicy](https://github.com/ben-manes/caffeine/blob/d084f2d3f1bb210dcba5b3f054039680178ef707/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/product/CaffeinePolicy.java), [CacheBuilderTest](https://github.com/ben-manes/caffeine/blob/master/guava/src/test/java/com/github/benmanes/caffeine/guava/compatibility/CacheBuilderTest.java), and [CacheBuilderFactory](https://github.com/ben-manes/caffeine/blob/master/guava/src/test/java/com/github/benmanes/caffeine/guava/compatibility/CacheBuilderFactory.java).

Stephan202:
> I suppose we can also remove some of the suppressions that are now irrelevant. I think that would be [CaffeinePolicy](https://github.com/ben-manes/caffeine/blob/d084f2d3f1bb210dcba5b3f054039680178ef707/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/product/CaffeinePolicy.java), [CacheBuilderTest](https://github.com/ben-manes/caffeine/blob/master/guava/src/test/java/com/github/benmanes/caffeine/guava/compatibility/CacheBuilderTest.java), and [CacheBuilderFactory](https://github.com/ben-manes/caffeine/blob/master/guava/src/test/java/com/github/benmanes/caffeine/guava/compatibility/CacheBuilderFactory.java).

Thanks for flagging. That actually uncovers that additional `@CanIgnoreReturnValue` annotations are necessary, because the `return self` cases (where `self` is casted) aren't flagged by `CanIgnoreReturnValueSuggester`. Doing another pass...

jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
```diff
@@ -552,6 +553,7 @@ private V removeNoCopyOrAwait(K key) {
   }
 
   @Override
+  @CanIgnoreReturnValue
   public boolean remove(K key, V oldValue) {
```
Stephan202, line 557: The annotation also applies to some other methods in this class; they're not flagged likely because there's no test calling them. I'll add a separate commit for this.

Stephan202 (commented):
Resolved most `@SuppressWarnings`; there remaining ones call a method whose return value should _generally_ not be ignored. One more commit coming up.

jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
ben-manes: Thanks. Mostly it is called by the jsr107 tck where our tests are to increase the code coverage due to possible gaps. The tck is a dependency that is unzipped during the build to run its tests, so a bit indirect and wouldn't be flagged.

Stephan202:
Updated the PR title and suggested commit message, since the PR now also covers some test cdoe.

ben-manes (approved):
I'll merge and release once @kluever gives his blessing.

ben-manes:
hey @kluever, sorry to bother you but can you please take a quick skim?

guava/src/main/java/com/github/benmanes/caffeine/guava/CaffeinatedGuavaLoadingCache.java
```diff
@@ -123,7 +123,7 @@ public V apply(K key) {
   }
 
   @Override
-  @SuppressWarnings({"CheckReturnValue", "FutureReturnValueIgnored"})
+  @SuppressWarnings("FutureReturnValueIgnored")
```
kluever, line 126: We generally have been recommending just capturing into a `var unused = ...` variable instead of using `@SuppressWarnings(...)`. Up to you tho.

kluever (commented):
LGTM

Do you want to add `@CheckReturnValue` to the package-info.java (or alternatively, on each top-level type?)

Stephan202:
> Do you want to add `@CheckReturnValue` to the package-info.java (or alternatively, on each top-level type?)

This was already done in ff1238567ede4001270ac6ed4682989b6b508c7e :)

Assuming each package has such a file, all were modified:
```sh
$ git grep -L CheckReturnValue -- '*package-info.java' | wc -l
0
```

Stephan202:
Some packages lack a `package-info.java`, but none seem to really need it (that said, happy to add :) )
```sh
$ git ls-files '*.java' -- ':!:*module-info.java' | xargs dirname | sort -u | while read d; do test -f "${d}/package-info.java" || echo "${d}"; done
caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache
caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/local
caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/node
caffeine/src/jmh/java/com/github/benmanes/caffeine
caffeine/src/jmh/java/com/github/benmanes/caffeine/cache
caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/impl
caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/sketch
caffeine/src/jmh/java/com/github/benmanes/caffeine/profiler
caffeine/src/test/java/com/github/benmanes/caffeine
caffeine/src/test/java/com/github/benmanes/caffeine/apache
caffeine/src/test/java/com/github/benmanes/caffeine/cache
caffeine/src/test/java/com/github/benmanes/caffeine/cache/buffer
caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues
caffeine/src/test/java/com/github/benmanes/caffeine/cache/stats
caffeine/src/test/java/com/github/benmanes/caffeine/cache/testing
caffeine/src/test/java/com/github/benmanes/caffeine/eclipse
caffeine/src/test/java/com/github/benmanes/caffeine/eclipse/acceptance
caffeine/src/test/java/com/github/benmanes/caffeine/eclipse/mutable
caffeine/src/test/java/com/github/benmanes/caffeine/google
caffeine/src/test/java/com/github/benmanes/caffeine/jsr166
caffeine/src/test/java/com/github/benmanes/caffeine/lincheck
caffeine/src/test/java/com/github/benmanes/caffeine/testing
examples/coalescing-bulkloader/src/main/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader
examples/coalescing-bulkloader/src/test/java/com/github/benmanes/caffeine/examples/coalescing/bulkloader
examples/write-behind-rxjava/src/main/java/com/github/benmanes/caffeine/examples/writebehind/rxjava
examples/write-behind-rxjava/src/test/java/com/github/benmanes/caffeine/examples/writebehind/rxjava
guava/src/test/java/com/github/benmanes/caffeine/guava
guava/src/test/java/com/github/benmanes/caffeine/guava/compatibility
jcache/src/test/java/com/github/benmanes/caffeine/jcache
jcache/src/test/java/com/github/benmanes/caffeine/jcache/configuration
jcache/src/test/java/com/github/benmanes/caffeine/jcache/copy
jcache/src/test/java/com/github/benmanes/caffeine/jcache/event
jcache/src/test/java/com/github/benmanes/caffeine/jcache/expiry
jcache/src/test/java/com/github/benmanes/caffeine/jcache/integration
jcache/src/test/java/com/github/benmanes/caffeine/jcache/management
jcache/src/test/java/com/github/benmanes/caffeine/jcache/processor
jcache/src/test/java/com/github/benmanes/caffeine/jcache/size
jcache/src/test/java/com/github/benmanes/caffeine/jcache/spi
simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/address/penalties
simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/gradient
simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/hill
simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/inference
simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/sim
simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/admission/bloom
```

guava/src/main/java/com/github/benmanes/caffeine/guava/CaffeinatedGuavaLoadingCache.java
Stephan202: I see 36 usages of `@SuppressWarnings("FutureReturnValueIgnored")`. Happy to make this change, but perhaps this is better done in a separate PR. Will await @ben-manes' verdict :)

ben-manes: I'd prefer `@SuppressWarnings` as other linters will flag unused variables (e.g. Eclipse). It's unfortunate as it does expand the suppression to the method instead of statement, but does a better job of avoiding warning spam. I'm open to revisiting to a better scheme separately so as to narrow the scope.

ben-manes:
> Some packages lack a package-info.java, but none seem to really need it (that said, happy to add :) )

If you'd like to add to those other packages that would be fine, I think only the `/main/` was my intent with a few missed. The javaPoet would be good as I forgot about it (codegen) and I didn't think tests / benchmarks / examples needed it but wouldn't care either way if added. Happy to do what you guys think is best 🙂

Stephan202:
Added another commit; as far as I can tell all packages now declare `@CheckReturnValue`.

I noticed that a subset of the existing packages also declare [nullability annotations](https://github.com/ben-manes/caffeine/blob/d084f2d3f1bb210dcba5b3f054039680178ef707/caffeine/src/main/java/com/github/benmanes/caffeine/cache/package-info.java#L38-L40) (as well as a copyright header, in some cases); I left that out for now. If I should e.g. add those things to just the packages in the `caffeine` module, lemme know.

ben-manes:
lgtm. We’re not using checker so it’s only for users, so probably unnecessary. You can merge once the ci passes 🙂

Stephan202:
:+1: Thanks for the quick review! Updated the suggested commit message to include this latest change.

ben-manes:
Released in 3.1.4

PR #1075, opened 2023-07-17
Add `Revved up by Gradle Enterprise` badge

clayburn:
This PR only gives the caffeine project the right to display the badge in this particular situation, as a badge in the project README.


runningcode (approved):

ben-manes:
I wasn't sure if this is something you'd want because the UI appears broken, making it look like GE is either unused or abandoned. In fact every CI build publishes a scan ([example](https://caffeine.gradle-enterprise.cloud/s/22kpo6pxzhb52)) and pushes to the remote cache, which both local and remote builds leverage. It has been a really great help in cutting the build times dramatically, but anyone who peeks at our dashboard wouldn't get that impression. If it is trivial to fix then I'd happily display the badge, else it seems like a toss-up. wdyt?

Issue #1347, opened 2023-11-29
Property "hibernate.javax.cache.uri" is ignored in JakartaEE deployments


mletenay:
**Situation:**
JakartaEE application with EAR file consiting of several EJB JAR modules, each with its own persistence context and its own hibernate 2nd level cache config.

**Problem:**
Migrating from EhCache 2.x to Caffeine, but as of 3.1.8 it does not work and completely ignores the _hibernate.javax.cache.uri_ specified in _persistence.xml_ descriptors.
The issue with "hibernate.javax.cache.uri" was already partially addressed in https://github.com/ben-manes/caffeine/issues/877 but that solution works only for Spring application.

As mention in https://github.com/ben-manes/caffeine/issues/877#issuecomment-1454981294 from security reasons, the uri loading was limited to **file:** and **classpath:** protocols.
The actual trouble is that in case of JakartaEE, the URI is neither of those but something like **jar:**file:/ejbmodule.jar!/META-INF/specific-cache.conf.

The fix is be trivial, just add jar: protocol handling to TypesafeConfigurator#resolveConfig, I will create PR for that ...

ben-manes:
Thank you! I’m hoping to get caught up over the weekend for this and the other two recent requests. Until this is released, you can use [setConfigSource](https://github.com/ben-manes/caffeine/blob/41cf7fdf7c0f7f62c502467f31d08cce0e43814b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/configuration/TypesafeConfigurator.java#L151) to resolve using your application strategy.

ben-manes:
I merged this a little early to base the unit test on top of. It took a little build magic to make an extra test resource jar that is placed on the classpath, so that the resource lookup resolves using the jar scheme. That uri is working so just need to duplicate the tests cases. I might get everything wrapped up this weekend for a release, but no promises.

mletenay:
I've looked at the unit tests when pushed my commit, but I am more junit/maven guy and your setup is slightly above my skills...
Anyway, thanks for nice library.

mletenay:
Sorry to bother you, but it would be nice to have a new release so we do not have to hack around this issue with custom call of seeConfigSource ...
Thanks a lot.

ben-manes:
[...]

PR #1348, opened 2023-11-29
#1347 - Fix for hibernate.javax.cache.uri in JakartaEE EARs

mletenay:
As mentioned in [#1347](https://github.com/ben-manes/caffeine/issues/1347), the URIs in case of JakartaEE jar files are neither **file:** nor **classpath:** but **jar:**.
Extend the TypesafeConfigurator#resolveConfig to properly pass also jar: protocol URLs, so the Caffeine can be used as Hibernate's 2nd level cache in complex JakartaEE applications.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/configuration/TypesafeConfigurator.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/configuration/TypesafeConfigurator.java
@@ -170,10 +170,6 @@ public final class TypesafeConfigurator {
       return ConfigFactory.defaultOverrides(classloader)
           .withFallback(ConfigFactory.parseFile(new File(uri), options))
           .withFallback(ConfigFactory.defaultReferenceUnresolved(classloader));
-    } else if (isResource(uri)) {
-      return ConfigFactory.defaultOverrides(classloader)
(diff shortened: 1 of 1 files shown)


jcache/src/main/java/com/github/benmanes/caffeine/jcache/configuration/TypesafeConfigurator.java
```diff
@@ -166,6 +170,14 @@ private static Config resolveConfig(URI uri, ClassLoader classloader) {
...
           .withFallback(ConfigFactory.parseFile(new File(uri), options))
           .withFallback(ConfigFactory.defaultReferenceUnresolved(classloader));
+    } else if ((uri.getScheme() != null) && uri.getScheme().equalsIgnoreCase("jar")) {
+        try (Reader reader = new InputStreamReader(uri.toURL().openStream())) {
+          return ConfigFactory.defaultOverrides(classloader)
+              .withFallback(ConfigFactory.parseReader(reader, options))
+              .withFallback(ConfigFactory.defaultReferenceUnresolved(classloader));
+        } catch (IOException e) {
+          throw new ConfigException.BadPath(uri.toString(), e.getMessage());
+        }
```
ben-manes, line 180: I think we could use `ConfigFactory.parseURL(uri.toURL(), options)`.

PR #1386, opened 2023-12-14
Set remote build cache to push only if authenticated

ribafish:
Updated remote build cache config to only push when authenticated -> if an unauthenticated push is attempted remote build cache will get turned off for the rest of the build. This can happen (does) on forked PR builds which run in an unsecure environment.


ben-manes (approved):
thanks!

PR #1518, opened 2024-02-09
Sieve2: a sieve that remembers

thomasmueller:
(Maybe you remember me, I helped a bit with the LIRS cache and the Bloom filter in the past).

I have seen the new "Sieve" cache algorithm and played a bit with it. I saw you have already implemented it, great! 

I find Sieve is quite fast, and can easily be made concurrent, without having to deal with segments and so on.

The problem is that it doesn't remember evicted keys, like LIRS and TinyLFU. I implemented a solution for that, in this Sieve2. It has a simple set-associative cache for keys that were evicted recently (you could use a sketch data structure for that... I just wanted something simple). This "memory" improves the hit rate a lot, and for some traces it is better than LIRS and the Caffeine cache. (Sometimes it is worse of course). The idea is: if an entry is added to the cache, and that entry was recently evicted, then it is added as a "hot" entry. (The size of the memory can be tweaked, and one could introduce multiple levels of "hotness", but it's probably best to keep it simple).

The PR is not polished at all: I don't expect that you merge it! I just wanted to let you know. Some of the changes are just to simplify testing; I guess there is a better way for that. I'll run the tests against a few policies and create an spreadsheet.


thomasmueller:
Random thoughts:
* Would it make sense to add a "meta policy" that predicts which is the best eviction policy (based on statistics), and then switches between the policies? The best data compression tools do that. There is a risk of slower operation, if not implemented carefully.
* I wonder if Sieve is patented, or if there is a plan for that.

ben-manes:
Neat, it sounds like you added a [victim cache](https://en.m.wikipedia.org/wiki/Victim_cache#:~:text=A%20victim%20cache%20is%20a%20hardware%20cache%20designed%20to%20decrease,cached%20in%20the%20victim%20cache.).

I like to use the ARC traces and the adaptive stress test (Corda large, 5x Lirs loop, Corda large) workloads, can you test those? It gives a good sense of real world patterns, e.g. the S3 and OLTP closely match the UMass web search and financial disk traces. The stress test is artificial but proves that the policy could work well in unknown situations and there’s no funny cherrypicking or parameter hacking. You can try larger traces by looking at the simulator wiki page for ones that interest you.

Caffeine tries to be competitive across all workload patterns. That means it has to explore a bit to self tune, which hurts the hit rates a little while booting them a lot. I like your coloring as giving equivalent scores if close; I use ~2% as a rule of thumb.

ben-manes:
I’ve seen a few meta policies, but never explored them. They use multi-armed bandits or similar to select the in use algorithm.

Sieve is a very tiny change from the classic Clock policy from Multics (1960s). I don’t think it would be considered novel enough, even though it changes the hit rate profile significantly by shifting from lru towards mru. But fundamentally it’s not very novel, just a nice twist.

ben-manes:
fwiw, wrt meta, Caffeine instead uses hill climbing (samples the hit rate) and jitter (admits 1% of warm candidates to unstick a hot victim).

maypok86:
It looks cool, but as far as I know the SIEVE authors tried to keep it as simple as possible. That's why they didn't add things like this, which obviously improve hit ratio.

maypok86:
By the way, you can call the authors here as well😄

thomasmueller:
@ben-manes I tried with Corda large, 5x Lirs loop, Corda large as follows:

`paths = [ "corda:trace_vaultservice_large.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "lirs:loop.trace.gz", "corda:trace_vaultservice_large.gz" ]`

and it was 25%. I can tweak it to reach 36% which is pretty close to the 39% of caffeine. Sieve and Cache2k have 0%; LIRS has 20%. I need to better understand what the loop trace is doing. I'll also test with the ARC traces.

I wonder, don't you check with different cache sizes? Testing with only one cache size seems problematic to me...

> it sounds like you added a victim cache

As far as I understand, a "victim cache" contains entries that are evicted due to hash collision... This is not what I added. Also, a "victim cache" also contains the values - in my case, it only contains the hash codes of the keys. Not the values. I think I added something very similar to the TinyLFU, but instead of using a sketch data structure or Bloom filter, I used an array of hash codes.

@maypok86 yes but an algorithm that is too simple might not be used at all. "Keep it as simple as possible, but not simpler". The original Sieve algorithm sounds like it's too simple to be competitive with very good, highly tweaked algorithms like Caffeine. Or rather, adding a little bit of complexity might make it competitive with LIRS and so on. But, possibly much faster, and still less complex.

maypok86:
@thomasmueller Totally agree, but the word "simpler" is literally in the title of the original paper. 🙃 In my opinion, the original version of SIEVE is better seen as a replacement for LRU when the developer doesn't want to bother with a more advanced algorithm. Caffeine takes care of implementing advanced algorithms and keeping them working correctly, which makes it actually ideal for the vast majority of Java programs (no need to worry about anything, everything works out of the box). But if a user has some very specific use case (or for example there are no suitable libraries in the language), because of which it was decided to write his own cache from scratch, it is not a fact that many people will want to invest in W-TinyLFU implementation. After all, in most cases people's knowledge of eviction policies is limited to LRU, ARC and CLOCK maybe.

ben-manes:
> I wonder, don't you check with different cache sizes? Testing with only one cache size seems problematic to me...

My goal with that test was to understand how the policy handled multiple extreme workload changes (LRU => MRU => LRU). Unfortunately those strongly biased traces are for very small caches so while I could chain them together, I don't have much larger traces that demonstrate the same extreme effects. If the cache is too tiny then there is not enough history for a policy to work from or if too large it quickly absorbs the entire workload for a perfect hit rate. The size of 512 had the effect that I was trying to achieve when exploring adaptive approaches, as some of my attempts got stuck or dampened the adaptivity after the first cycle. It also produced a really neat chart for how the internals were responding.

<details><summary>Adaptivity</summary>

image
</details>

I am very open to more traces, mixed workloads, different sizes, etc. I don't think there are correct and incorrect traces, golden workloads, that workload patterns age out to be irrelevant, and dislike these types of excuses to justify cherrypicked results. I merely have a few preferred traces that show a variety of real world patterns or challenging behaviors and at a glance let me make a rough assessment. 
 
> This is not what I added.

oh yes, sorry, I had skimmed too quickly! 😄 
You made a nice way to capture the history and respond to it. I like how compact and efficient you made it.

ben-manes:
Here are a few workloads using your algorithm! Below are the charts and commands. I added Tencent Photo since that's one that the Sieve authors favor.

I ran with the [optimal policy](https://github.com/ben-manes/caffeine/blob/master/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/opt/ClairvoyantPolicy.java) which makes the simulation memory hungry and slow. There is probably a smart way to use an embedded db to avoid and avoid recomputing between runs. It is currently an `O(n lg m)` on-heap algorithm, where `n` is the trace length and `m` is the cache size.

Someday, I'd like to further improve Caffeine's hit rates. Primarily, I think that the current hill climber is a bit too naive and we can gain a bit using a more aggressive climber. For example, we currently observe over a large sample period to avoid noise and decay the step size for convergence. That could start with a small sample period that grows over time, so it makes rapid adjustments early on in large traces with later fine tuning. It also only remembers a two observation periods whereas a few more might lead to better choices. Regardless, we're pretty competitive and I haven't seen any novel new approaches that outpaces anything else, e.g. 10pt gap, so we're stuck trying to get small returns with minimal harm.

<details open><summary>DS1</summary>

ds1

| Policy           | 100,000 | 200,000 | 300,000 | 400,000 | 500,000 | 600,000 | 700,000 | 800,000 |
| ---------------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- |
| irr.Lirs         | 12.40   | 15.55   | 25.08   | 34.69   | 44.27   | 53.15   | 60.99   | 67.64   |
| linked.Sieve     | 10.12   | 17.93   | 23.94   | 29.36   | 37.51   | 46.27   | 55.40   | 63.76   |
| linked.Sieve2    | 12.04   | 20.37   | 26.19   | 31.86   | 38.68   | 46.84   | 55.40   | 63.52   |
| opt.Clairvoyant  | 25.42   | 39.79   | 50.92   | 59.96   | 67.09   | 72.97   | 77.57   | 81.27   |
| product.Cache2k  | 9.90    | 18.89   | 27.67   | 36.20   | 44.46   | 52.24   | 58.69   | 65.01   |
| product.Caffeine | 11.63   | 22.95   | 33.39   | 42.51   | 50.95   | 58.69   | 64.82   | 70.17   |

```console
JAVA_VERSION=21 gradle simulator:simulate -q \
  -Dcaffeine.simulator.files.paths.0="arc:/Users/ben/Documents/traces/arc/DS1.lis.xz" \
  --maximumSize=1_000_000,2_000_000,3_000_000,4_000_000,5_000_000,6_000_000,7_000_000,8_000_000 \
  -Dcaffeine.simulator.policies.0=opt.Clairvoyant \
  -Dcaffeine.simulator.policies.1=irr.Lirs \
  -Dcaffeine.simulator.policies.2=linked.Sieve \
  -Dcaffeine.simulator.policies.3=linked.Sieve2 \
  -Dcaffeine.simulator.policies.4=product.Cache2k \
  -Dcaffeine.simulator.policies.5=product.Caffeine \
  -PjvmArgs=-Xmx8g \
  --title=DS1
```
</details>

<details><summary>S3</summary>

s3

| Policy           | 100,000 | 200,000 | 300,000 | 400,000 | 500,000 | 600,000 | 700,000 | 800,000 |
| ---------------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- |
| irr.Lirs         | 12.40   | 15.55   | 25.08   | 34.69   | 44.27   | 53.15   | 60.99   | 67.64   |
| linked.Sieve     | 10.12   | 17.93   | 23.94   | 29.36   | 37.51   | 46.27   | 55.40   | 63.76   |
| linked.Sieve2    | 12.04   | 20.37   | 26.19   | 31.86   | 38.68   | 46.84   | 55.40   | 63.52   |
| opt.Clairvoyant  | 25.42   | 39.79   | 50.92   | 59.96   | 67.09   | 72.97   | 77.57   | 81.27   |
| product.Cache2k  | 9.90    | 18.89   | 27.67   | 36.20   | 44.46   | 52.24   | 58.69   | 65.01   |
| product.Caffeine | 11.63   | 22.95   | 33.39   | 42.51   | 50.95   | 58.69   | 64.82   | 70.17   |

```console
JAVA_VERSION=21 gradle simulator:simulate -q \
  -Dcaffeine.simulator.files.paths.0=arc:/Users/ben/Documents/traces/arc/S3.lis.xz \
  --maximumSize=100_000,200_000,300_000,400_000,500_000,600_000,700_000,800_000 \
  -Dcaffeine.simulator.policies.0=opt.Clairvoyant \
  -Dcaffeine.simulator.policies.1=irr.Lirs \
  -Dcaffeine.simulator.policies.2=linked.Sieve \
  -Dcaffeine.simulator.policies.3=linked.Sieve2 \
  -Dcaffeine.simulator.policies.4=product.Cache2k \
  -Dcaffeine.simulator.policies.5=product.Caffeine \
  -PjvmArgs=-Xmx8g \
  --title=S3
```
</details>

<details><summary>OLTP</summary>

oltp

| Policy           | 250   | 500   | 750   | 1,000 | 1,250 | 1,500 | 1,750 | 2,000 |
| ---------------- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- |
| irr.Lirs         | 18.27 | 26.87 | 31.71 | 34.82 | 37.24 | 39.20 | 40.93 | 42.52 |
| linked.Sieve     | 11.79 | 20.65 | 27.45 | 32.54 | 35.16 | 37.09 | 38.90 | 41.18 |
| linked.Sieve2    | 22.87 | 32.25 | 36.96 | 39.99 | 42.15 | 43.91 | 45.34 | 46.58 |
| opt.Clairvoyant  | 38.47 | 46.43 | 50.67 | 53.62 | 55.84 | 57.62 | 59.13 | 60.40 |
| product.Cache2k  | 19.35 | 27.86 | 32.81 | 36.40 | 39.14 | 41.25 | 43.02 | 44.45 |
| product.Caffeine | 23.81 | 33.15 | 37.59 | 39.63 | 41.87 | 43.64 | 44.86 | 45.93 |

```console
JAVA_VERSION=21 gradle simulator:simulate -q \
  -Dcaffeine.simulator.files.paths.0=arc:/Users/ben/Documents/traces/arc/OLTP.lis.xz \
  --maximumSize=250,500,750,1000,1250,1500,1750,2000 \
  -Dcaffeine.simulator.policies.0=opt.Clairvoyant \
  -Dcaffeine.simulator.policies.1=irr.Lirs \
  -Dcaffeine.simulator.policies.2=linked.Sieve \
  -Dcaffeine.simulator.policies.3=linked.Sieve2 \
  -Dcaffeine.simulator.policies.4=product.Cache2k \
  -Dcaffeine.simulator.policies.5=product.Caffeine \
  -PjvmArgs=-Xmx8g \
  --title=OLTP
```
</details>

<details><summary>Tencent Photo</summary>

tencentPhoto

| Policy           | 200,000 | 400,000 | 600,000 | 800,000 | 1,000,000 | 1,200,000 | 1,400,000 | 1,600,000 |
| ---------------- | ------- | ------- | ------- | ------- | --------- | --------- | --------- | --------- |
| irr.Lirs         | 46.09   | 54.50   | 59.00   | 61.88   | 63.92     | 65.46     | 66.68     | 67.68     |
| linked.Sieve     | 47.71   | 55.91   | 60.28   | 63.05   | 65.10     | 66.59     | 67.85     | 68.88     |
| linked.Sieve2    | 49.75   | 57.43   | 61.47   | 64.09   | 65.97     | 67.41     | 68.55     | 69.49     |
| opt.Clairvoyant  | 62.39   | 68.27   | 71.29   | 73.24   | 74.61     | 75.69     | 76.54     | 77.20     |
| product.Cache2k  | 45.22   | 53.26   | 57.61   | 60.46   | 62.53     | 64.14     | 65.44     | 66.52     |
| product.Caffeine | 49.36   | 55.82   | 60.40   | 62.74   | 64.52     | 65.09     | 66.38     | 65.80     |

```console
JAVA_VERSION=21 gradle simulator:simulate -q \
  -Dcaffeine.simulator.files.paths.0=lirs:/Users/ben/Documents/traces/libcachesim/tencentPhoto.trace.xz \
  --maximumSize=200_000,400_000,600_000,800_000,1_000_000,1_200_000,1_400_000,1_600_000 \
  -Dcaffeine.simulator.policies.0=opt.Clairvoyant \
  -Dcaffeine.simulator.policies.1=irr.Lirs \
  -Dcaffeine.simulator.policies.2=linked.Sieve \
  -Dcaffeine.simulator.policies.3=linked.Sieve2 \
  -Dcaffeine.simulator.policies.4=product.Cache2k \
  -Dcaffeine.simulator.policies.5=product.Caffeine \
  -PjvmArgs=-Xmx16g \
  --title="Tencent Photo"
```
</details>

ben-manes:
Closing but always happy to keep discussing!

thomasmueller:
> Closing but always happy to keep discussing!

Yes, I would also close it. I just wanted to let you know that I'm looking into that.

I was playing with this a bit more in the last days, but so far couldn't beat the Caffeine hit rate. I now want to plug in the TinyLFU admittor into Sieve (basically replace LRU with Sieve), and see if that works. I think it would be faster than LRU, doesn't require segments for concurrency, and there is a possibility to set the "visited" flag -- this is something LRU doesn't have. The temporal admission window could also be Sieve, or possibly an even faster algorithm, e.g. direct mapped cache.

ben-manes:
I haven't seen any newer eviction policies that make a leap ahead of the competition in any workload, even if at the expense of others. Perhaps naively I feel like we've plateaued, perhaps because many algorithms are rediscoveries of the same insights. The ML ones are interesting as newer techniques, but impractically expensive training and not yet showing a strong benefit. I suppose my thinking is we need another out-of-the-box insight, like Gil has with TinyLFU, from which us engineers refine to make practical.

The gap between the best algorithms and optimal is pretty large in some of those traces (S3, OLTP, Tencent). Those are disk-based which makes me wonder about prefetching, as its a strategy that we can't easily exploit in the general case. Those are likely linear scans where detecting a pattern could assist or a key might belong to a clusters of related content that is likely to be requested next. It's hard for us since we only have abstract keys, but then those non-resident ghost queues are just that but shallow. So I kind of wonder if there would be a way to cluster keys to preload or evict groups, in the hope that this meta-layer could make good predictions to improve cache efficiency.

Kind of vague rambling, sorry...

thomasmueller:
> I haven't seen any newer eviction policies that make a leap ahead of the competition in any workload

The hit-rate: no. However, it's still possible to improve performance, I think. (And possibly reduce complexity.)

> makes me wonder about prefetching

Yes [prefetching](https://en.wikipedia.org/wiki/Cache_prefetching) is interesting! The idea is very old, but mostly only linear scans are supported in reality; mostly CPU level and file system level. For the application level, people think on how to do that from time to time, e.g. https://issues.apache.org/jira/browse/OAK-6571 :-) but then rarely do this. Possibly if there are some tools that could help, it might be easier. Keys then can not be abstract then.

1a1a11a:
Interesting work @thomasmueller! 
We kept the policy simple, but we also show it is possible to use it as a primitive to construct advanced algorithms, just as you have done. I tried S3-SIEVE (replacing the main FIFO with SIEVE in S3-FIFO, it is more robust than SIEVE, but also harder to implement. There might be other ideas, we would love to see these ideas. 

Re: meta policy, see [LeCaR](https://www.usenix.org/system/files/conference/hotstorage18/hotstorage18-paper-vietri.pdf) and [CACHEUS](https://www.usenix.org/conference/fast21/presentation/rodriguez). The problem is overhead and complexity, but ideas are interesting. 

@ben-manes 
Caffeine is great, however, there are certain issues with adaptive policy 
> I think that the current hill climber is a bit too naive and we can gain a bit using a more aggressive climber. 
As you mentioned, adaptive policies have many parameters, like how fast you adapt and how large the window should be. For example, by adding a jitter with magic parameters, Caffeine can be better than SIEVE on certain workloads, as you showed [this issue](https://github.com/Yiling-J/theine/issues/21). Many designs have been tuning on a set of traces (like using the training data for evaluation in ml), but when evaluated at a large scale (>6000 traces), adaptive algorithms fail to behave better than static ones. Of course, I am not saying static algorithms are always better, but at least they are easy to tune and understand. Maybe one day, there will be a better adaptive algorithm, but so far, I have not seen it. 

About the gap between Belady and state-of-the-art. As we all have observed, the miss ratios of many of the state-of-the-art algorithms have converged (show similar miss ratios) on many workloads, __I doubt whether the gap can be bridged__. As an online system, it has limited information to learn (and adapt); there might be a bound limited by information theory. 

Happy to chat more if anyone is interested.

1a1a11a:
@thomasmueller If you want to add SIEVE2 to libCacheSim (https://libcachesim.com), I am happy to merge it so that other researchers are aware of your work.

ben-manes:
I think Caffeine's climber can be better, but that's so far been a personal opinion and not data driven. I think part of the reason that I feel that way is ours is extremely simple, yet surprisingly robust and effective. Each time that I think there might be a flaw to investigate it turns out to be something else and the climber worked pretty well. I don't think static algorithms are practical for the majority of developers who won't have the knowledge or awareness to make good upfront decisions. We see that in database tuning, for example, where a light usage of ML can solve the configuration explosion problem. Unfortunately I think our eviction policy might be an outlier as tuning other algorithms seems to be less straightforward, not as effective, or incurs a high overhead.

But yeah, I think we're hitting a peak at hit rates from an admission/eviction policy. Making it less complex is worthwhile, just not as interesting for me due to a sunk cost fallacy of having working code. I was wondering if evicting a cluster of keys could be of a benefit, e.g. if a correlated load, but likely the individual metadata is enough to infer their lack of reuse.

So topics where I find myself drawn to are,
1. Optimizing for other metrics, like being latency-aware to improve overall system performance. I had some ideas on this and discussed it with others, but there is very little data to experiment with. That makes me a bit cautious because there is a not much variety of workload types from what data is available, so any promising solution might be overfit.
2. Making application prefetching more realizable. There is a lot of patterns and clustering of data that could be computed offline to make smart, online, optimistic loads. In the past a photo catalog might benefit from a linear scan if locally defragmented, but a cloud catalog may have poor data placement by having those images scattered across remote disks due to key hashing. I think this is too vague to really tackle, but there is probably a lot of low hanging gains because explicit preloading is rare and there are many cases where being predictive could improve the user experience.
3. Popularize [request coalescing](https://github.com/ben-manes/caffeine/tree/master/examples/coalescing-bulkloader-reactor) to hide latency. I think the tooling is all there, it just isn't used very broadly so needs some evangulism.

thomasmueller:
> Making it less complex is worthwhile, just not as interesting for me due to a sunk cost fallacy of having working code.

I completely understand this!

> Optimizing for other metrics

One "easy" metric, for me, is speed. That's why I think Sieve is interesting. Let's assume with a high hit rate, and we have concurrency. With LRU, we need to use segments, but concurrency issues remain up to some point. With Sieve, we don't need to use segments, and a hit is mostly just a a lookup and setting a flag.

In my fork I have added FeedbackWindowSieveTinyLfuPolicy which is a copy of FeedbackWindowTinyLfuPolicy :  https://github.com/ben-manes/caffeine/commit/9e24b65e9fe12010145a458b82dd7a4091b72b7e  performance (single threaded) is basically the same as with the old one, and hit rate is also the same.

Screenshot 2024-02-22 at 12 04 42

In the (existing) code I do see some challenges:
* I don't fully understand what the "probation" part is - it seems a bit under-documented.
* There seem to be a couple of hacks, like using headProbation.prev.prev
* I do need to set node.visited = true when adding to the window area, which seem weird.

But that's about it. If I have time, I think I will continue with this. Specially, the multi-threaded performance aspect seems interesting to me.

ben-manes:
> With LRU, we need to use segments, but concurrency issues remain up to some point. 

Actually segments are a bad approach because while the data is uniformly distributed, the access distribution is heavily skewed (zipfian). That's why segment locks don't scale well as the hot entries incur contention. Caffeine and its predecessors (CLHM, if you recall our first discussions!) solve this by decoupling the access event from the LRU lock, so we can record and replay events. We use a j.u.c.LongAdder inspired dynamic striping of ring buffers (vs counters). That approach achieves 1/3rd the throughput of an unbounded ConcurrentHashMap, e.g. 940M reads/s on my new macbook.

> With Sieve, we don't need to use segments, and a hit is mostly just a a lookup and setting a flag.

Yes, that's because it is a Clock-based algorithm, that's an inherited feature and not unique. Clock has been popular since the 1960s, originally because it avoided data movement by the page replacement by replacing the entry at the hand. It could linearly scan over pages resetting the mark bit until a victim was found, which make it very efficient given the hardware of the time. That became pretty standard and extended as "Generalized Clock" for an m-bit counter, e.g. Postgres uses 2-bits iirc for a 0-4 range, meaning if may require 4 full scans in the worst case. Clock-based is probably the most popular algorithm in performance critical areas like databases, operating systems, etc for the historical properties, and then the concurrency benefit in the modern era. Linux for example uses the Double Clock policy, which incorporates non-resident entries.

When I write quick-and-dirty caches then Clock-based is my go-to since it is trivial to maintain, highly concurrent, and easy to adapt (like GClock or Sieve). That's what the early versions of CLHM used, since it was a pragmatic choice, and that project was originally meant for personal research rather than end users. I wanted to find a way to decouple the eviction policy from the concurrency scheme to avoid it limiting the design decisions and incurring an O(n) worst eviction. While it is easy to mitigate that in GClock by using a scan threshold, it's less clear for others like Double Clock, ClockPro, CAR. When you are thinking about tail latencies then predictable behavior is preferred instead of random hiccups or sudden drops in the hit rate.

> In my fork I have added FeedbackWindowSieveTinyLfuPolicy which is a copy of FeedbackWindowTinyLfuPolicy

Sorry, `FeedbackWindowTinyLfuPolicy` was a somewhat failed experiment towards adaption. I think in that version I tried to use a bloom filter as a feedback mechanism, aka like ARC does with its ghost queues. It worked for the first adjustment but not if chaining traces of different workload patterns. The `HillClimberWindowTinyLfuPolicy` with the `SimpleClimber` is what worked out the best for me and is similar to Caffeine's internal logic.

> I don't fully understand what the "probation" part is - it seems a bit under-documented.

That's Segmented LRU ([paper](https://github.com/ben-manes/caffeine/files/14380728/SLRU.-.Caching.strategies.to.improve.disk.system.performance.pdf)), an early disk policy based on the observation that most items are one-hit wonders. Since they pollute an LRU cache, the authors used a smaller probationary region before promoting to a protected region. This was an easy means of mixing frequency and recency for an O(1) policy, and that insight led to many subsequent ones with similar a structure and own name for those regions. Here it is used for the "main" region because we want to select a victim with low recency and frequency, and quickly discard if TinyLFU admitted it prematurely.

> There seem to be a couple of hacks, like using headProbation.prev.prev

I was definitely just iterating on ideas instead of code quality. I think the hill climber may be a little more polished.

> I do need to set node.visited = true when adding to the window area, which seem weird.

🤷 

> If I have time, I think I will continue with this. Specially, the multi-threaded performance aspect seems interesting to me.

I think TinyLFU is less multi-threaded friendly due to the sketch increments. I'm sure those could be atomic counters and dissipate as hotspots due to saturation. I never tried optimizing for that since I had a higher-level scheme. The version in Caffeine is optimized for the cpu cache line, whereas the simulator's is the slightly older version that you helped me write.

P.S. Actually multi-threading was what drew me into caching and not the policy. I had implemented a Clock to replace synchronized LRUs to fix performance problems at work and became curious as to whether a true LRU could be made concurrent (hence CLHM). I did that as picking an impossible problem for a motivational scenarios to learn concurrent algorithms, e.g. elimination stacks. That anything worked out or anyone used my code was a surprise.

thomasmueller:
Thanks a lot for your insights! Very nice read, lots of details!

> O(n) worst eviction... scan threshold...

This can become a problem for Sieve as well I assume. I'll have to do some reading / research here.

> FeedbackWindowTinyLfuPolicy was a somewhat failed experiment

Ups, then I picked the wrong one :-)

> I think TinyLFU is less multi-threaded friendly due to the sketch increments.

Yes, that's what I want to work on as well. If the sketch increments are more localized (e.g. in a cache line), then it would reduce CPU cache misses, and and might help and simplify concurrency. How to do this exactly needs some more work. It would be similar to a blocked Bloom filter / fuse filter.

> P.S. Actually multi-threading was what drew me into caching and not the policy. 

Interesting! Thanks a for sharing!

ben-manes:
Oh, it occurred to me that you are trying to add Sieve to SLRU, instead of replacing it. Basically "Sieve is to Clock" as "SLRU is to LRU" by giving an item a shorter duration to prove its worth. So you can replace the entire "main" region with Sieve.

Setting the `visited` on insertion into the "window" region makes sense as you are giving new entries a longer duration. That's our probationary space for new arrivals, so they need a little time to build up usage and prove their worth. Otherwise they would be evicted prematurely, so the hit rate drops. You should use Clock here instead of Sieve.

The hill / feedback climbers adapt the size of the regions, thereby giving more time or less to new arrivals to justify their admission into the main region. If the workload shows itself to have strong MRU behavior, like a sql row scan, then it shrinks the window region to more aggressively evaluate the item's worth. On the other hand if the workload has LRU behavior, like a message broker or job scheduler, then it grows the window to give arrivals a longer time to be retained. In that case the recency is a strong signal so the LRU victim of a large window region has aged out and may not be worth admitting by its frequency alone. Its popularity score would quickly fall due to TinyLFU's reset period, so even if accepted prematurely then the main region would aggressively drop it if unused (SLRU / Sieve).

`Clock | TinyLFU | Sieve` with adaptive region sizing is what you're trying for. Then make the sketch use atomics and it should have pretty decent concurrency.

thomasmueller:
> So you can replace the entire "main" region with Sieve.

OK I'll try that.

> Setting the visited on insertion into the "window" region makes sense as you are giving new entries a longer duration.

I see. BTW now with HillClimberWindowTinyLfuPolicy, it is less important to set visited = true. I mean, less important than with FeedbackTinyLfuPolicy, where it was very important, in order to get an acceptable hit rate. But it still is better: the hit rate is slightly better.

> Then make the sketch use atomics and it should have pretty decent concurrency.

Yes.

I think if done correctly (and unless if I misunderstand something), then the Linux buffer cache could use a "Sieve (+ Window) + TinyLFU" variant. AFAIK Linux currently uses a ClockPro variant (need to read on that some more.) I see you have a ClockProPlusPolicy variant, and that has a very good hit rate as well... so even if possible, it would be quite hard... Improvements to the TinyLFU implementation would be needed. But it's a challenge!

ben-manes:
> AFAIK Linux currently uses a ClockPro variant (need to read on that some more.)

It is similar but very different. I tried my best to reimplement it from their docs, see [DClock](https://github.com/ben-manes/caffeine/blob/master/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/DClockPolicy.java). They modify the algorithms fairly often so it may be out of date or have mistakes as not verifiable. Since then they introduced a multi-generational LRU, which iirc was like S4LRU except used lg(n) tiers. In the commit history they discuss an PID controller to adapt region sizes, which is something that I've looked into but never tried. It's worth glancing over.

PR #1550, opened 2024-02-25
When `AsyncCacheLoader.asyncLoadAll` returns excess keys, wait for all keys to be added to the cache before returning requested keys

### LINKED ISSUE #1409: Unnecessary fetches when using asynchronous bulk load
For a cache backed by AsyncCacheLoader that 
- loads values for keys that were not specifically requested ("pre-fetch")
- and where getAll() is called in close successions.

**Observed behavior:**

Depending on timing, asyncLoadAll() may sometimes be called unnecessarily for already pre-fetched items.

A reproducer can be provided if needed (that injects artificial delays in the map returned by asyncLoadAll()).
The following notes on LocalAsyncCache should however explain the behavior:

1) Completion of getAll() depends only on items that were actually requested
```
 return composeResult(futures);
```
https://github.com/ben-manes/caffeine/blob/v3.1.8/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalAsyncCache.java#L149

2) AsyncBulkCompleter completes futures in the following order - first requested items, then pre-fetched items:
```
        fillProxies(result);
        addNewEntries(result);
```
https://github.com/ben-manes/caffeine/blob/v3.1.8/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalAsyncCache.java#L260

The above results in behavior subject to timing between completion of addNewEntries(result) and when the next getAll() is triggered.
 
If getAll() is triggered again after the Future returned by a previous getAll() completed, but before "pre-fetched" items are added to the cache by addNewEntries(result), then the second call of getAll() could re-trigger a fetch of items that were already fetched (but not yet added to the cache.

Assumption is that getAll() completion immediately after requested items are added to the cache (by ```fillProxies(result)```) is an optimization.

For a scenario where asyncLoadAll() might be very expensive (as it is in my use case), this optimization is not the appropriate trade-off.

**Question**:

Would it be possible to add an option that allows completion of getAll() only after all fetched items have been added to the cache? 
The effect of this should be the same as swapping the order of   ```fillProxies(result); ``` and   ```addNewEntries(result); ``` in AsyncBulkCompleter.

allanrodriguez:
Fixes #1409.

Verified by running the `getAll_exceeds` test with the `ForkJoinPool` executor. Running this test configuration before the change would fail sporadically due to a race between the excess entries being added to the cache and the cache size assertion.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalAsyncCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalAsyncCache.java
@@ -144,8 +144,7 @@ interface LocalAsyncCache<K, V> extends AsyncCache<K, V> {
     try {
       var loader = mappingFunction.apply(
           Collections.unmodifiableSet(proxies.keySet()), cache().executor());
-      loader.whenComplete(completer);
-      return composeResult(futures);
+      return loader.whenComplete(completer).thenCompose(ignored -> composeResult(futures));
     } catch (Throwable t) {
       completer.accept(/* result */ null, t);
       throw t;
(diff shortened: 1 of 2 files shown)


ben-manes:
looks perfect, thanks!

PR #1688, opened 2024-05-05
Adding Bootique to the list of projects that work with Caffeine

andrus:
A specific Bootique/Caffeine example is provided here: https://github.com/bootique-examples/bootique-jcache-examples/tree/3.x/jcache-caffeine-example


ben-manes:
Can you make it flow with the rest of the rendered text? You'll notice that it is shaped, as ragged text is harder to skim. Probably the shorter "A minimally opinionated framework" is enough to not look like an outlier, and can then be ordered to maintain the triangular form.

andrus:
Done.

PR #1695, opened 2024-05-20
Add `caffeine-coroutines` in README

be-hase:
Hello. I always use Caffeine.

I have implemented a library called [caffeine-coroutines](https://github.com/be-hase/caffeine-coroutines), which adds Kotlin coroutines support to Caffeine. If possible, could you add this to the README?

There is also a Kotlin wrapper called [Aedile](https://github.com/sksamuel/aedile), but for the following reasons, I prefer to provide extensions rather than a wrapper.

https://github.com/be-hase/caffeine-coroutines?tab=readme-ov-file#we-will-primarily-focus-on-coroutine-support

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Use Caffeine in a community provided integration:
 * [ScalaCache][scala-cache]: Simple caching in Scala
 * [JHipster][jhipster]: Generate, develop, deploy
 * [Aedile][aedile]: Kotlin wrapper for Caffeine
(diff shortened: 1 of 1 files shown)


ben-manes:
Sure, but can you make the item length shorter and align into a triangular flow? It’s a bit wordy in comparison

ben-manes:
thanks, merged in manually

be-hase:
thanks. :)

PR #1789, opened 2024-10-26
Fix incorrect date format pattern in unit test

### LINKED ISSUE #1788: `Issue30Test` uses incorrect date time format pattern
https://github.com/ben-manes/caffeine/blob/0eaf6f21983a042073c89e99a3309dc13fbeadf5/caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues/Issue30Test.java#L154

- `h` is 'clock-hour-of-am-pm (1-12)', without am/pm marker (`a`) this might not be intended
Should maybe be `H` ('hour-of-day (0-23)')?
- `M` is 'month-of-year'
Should probably be `m` ('minute-of-hour')

So the pattern should probably be: `HH:mm:ss.SSS`

If this won't require signing the CLA, I can submit a pull request for this. Otherwise it might be easier if you fix this incorrect pattern.

Marcono1234:
Fixes #1788

Regarding the change `h` ('clock-hour-of-am-pm (1-12)') to `H` ('hour-of-day (0-23)'):
Maybe the usage of `h` was intended before, since it is also using the `US` locale, however then maybe it should have included the am/pm marker?

### DIFF
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues/Issue30Test.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/issues/Issue30Test.java
@@ -151,7 +151,7 @@ public final class Issue30Test {
 
   static final class Loader implements AsyncCacheLoader<String, String> {
     private static final DateTimeFormatter FORMATTER =
-        DateTimeFormatter.ofPattern("hh:MM:ss.SSS", US);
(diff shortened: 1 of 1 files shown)


ben-manes:
The static analyzers prefer an explicit locale. This is just for debugging the test so I think it’s okay. Thanks!

PR #1795, opened 2024-11-06
Use the ToolProvider SPI to run google-java-format

cushon:
This makes the compile-time dependency on google-java-format optional.

### DIFF
--- a/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/LocalCacheFactoryGenerator.java
+++ b/caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/LocalCacheFactoryGenerator.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.NavigableMap;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.spi.ToolProvider;
 import java.util.stream.Stream;
 
 import com.github.benmanes.caffeine.cache.local.AddConstructor;
@@ -57,8 +58,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import com.google.common.io.Resources;
-import com.google.googlejavaformat.java.Formatter;
-import com.google.googlejavaformat.java.FormatterException;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.ParameterizedTypeName;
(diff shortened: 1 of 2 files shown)


ben-manes:
thanks! Is there a strong reason to prefer the SPI or is it just a nicer practice?

I can polish to pass the CI and switch to a runtime dependency to merge this. I'm not sure why it recompiled in each job instead of using the build cache, so sorry for that warning spam.

cushon:
> thanks! Is there a strong reason to prefer the SPI or is it just a nicer practice?
> 
> I can polish to pass the CI and switch to a runtime dependency to merge this. I'm not sure why it recompiled in each job instead of using the build cache, so sorry for that warning spam.

Thanks for the lightning fast review!

The motivation for this was an internal bazel build for google-java-format. The formatter's tests depend on caffeine, which depends on google-java-format, so I was having to rebuild caffeine when testing some formatter changes. So not a common use-case, but I also didn't see downsides to using the SPI here. Thanks again!

ben-manes:
> So not a common use-case...

haha, well I know that fun. Caffeine's build depends on errorprone which depends on caffeine, and Gradle rewrites that to a project dependency so it relies on the pre-built classes. I have to [trick it](https://github.com/ben-manes/caffeine/blob/8b835c96ae6b1ea64b0706bc0484291c0c6f93f3/gradle/plugins/src/main/kotlin/quality/errorprone.caffeine.gradle.kts#L23-L24) to include the jar.

ben-manes:
> The formatter's tests depend on caffeine, which depends on google-java-format

oh can you explain this? The published metadata, [pom](https://central.sonatype.com/artifact/com.github.ben-manes.caffeine/caffeine) and [module](https://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/3.1.8/caffeine-3.1.8.module), don’t include that as a dependency. It should only be a build phase in this repository and not in a user’s, so it sounds like the thirdparty java_import rule might be excessive?

cushon:
> > The formatter's tests depend on caffeine, which depends on google-java-format
> 
> oh can you explain this? The published metadata, [pom](https://central.sonatype.com/artifact/com.github.ben-manes.caffeine/caffeine) and [module](https://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/3.1.8/caffeine-3.1.8.module), don’t include that as a dependency. It should only be a build phase in this repository and not in a user’s, so it sounds like the thirdparty java_import rule might be excessive?

The key part here was that this is an internal bazel build, which is building everything from source and not relying on the published artifacts. So it's probably _really_ not a common use-case, but it's helpful being able to drop the build-time dep for that.

ben-manes:
great, I figured something like that and likely for reproducible builds to protect against code injection attacks. Thanks for following up.

Issue #594, opened 2021-09-13
remove @Nullable from LoadingCache.get()


cvmocanu:
Currently `LoadingCache.get()` is marked with `@Nullable`.
This is incorrect, because Caffeine can't know the nullability of the cache loader function.

This causes a lot of trouble in actual usage, when you have a cache loader that never returns null (e.g. it's a simple `new MyClass()` or some other factory call that never returns null).
Because of the `@Nullable` on the `LoadingCache.get()` method, IntelliJ marks all usages of the result of the `get()` method as potential NPEs.
As a result, we either have to sprinkle the code with asserts or with suppress-warnings, neither of which is really required, because the `get()` method will never return null.

The annotation should be removed, because Caffeine can't know if the result is nullable or not.

ben-manes:
What is the correct way to mark this method to satisfy IntelliJ? If we remove the annotation then the defaults from package-info is that it is non-null, which may be incorrect.  

https://github.com/ben-manes/caffeine/blob/15330a16b509b56ae32a9bd0cb703e36ff6caa1f/guava/src/main/java/com/github/benmanes/caffeine/guava/package-info.java#L24

I don't think there is a correct way to annotate this method, since unlike `@PolyNull` on `get(key, mappingFunction)` we can't peek into the `CacheLoader`. I am fine with doing the least bad option, though. What would [JSpecify](https://github.com/jspecify/jspecify) recommend in this case  @cpovirk?

/cc @DavyLandman @vlsi

vlsi:
1. `Map#get` is often quoted when arguing on null safety. I think `@Nullable ... get()` is the best we can have with the current tooling.

    See the relevant discussions in https://github.com/jspecify/jspecify/issues/2

2. However, `LoadingCache#get()` seems to be a bit different: `LoadingCache#get()` seems to return `null` if and only if the loader returns `null`, so if the loader return type is non-nullable, then `LoadingCache#get()` should be non-nullable.  So the following might be slightly better (I don't remember if that signature has been discussed earlier):

```java
public interface LoadingCache<K extends Object, V extends @Nullable Object> extends Cache<K, V> {
  V get(K key);
```

    What it means is `V` can be any type, including `nullable` or `non-nullable Object`. If client declares `LoadingCache<String, Integer>`, then the type of `get()` would be non-nullable `Integer`. If the client declares `LoadingCache<String, @Nullable Integer>`, then the type of the getter becomes nullable.

[...]

PR #1805, opened 2024-12-03
Avoid giving the result of loads a non-null type.

cpovirk:
https://github.com/ben-manes/caffeine/commit/2542ca8ca22fe4c30b62ee569ee71436fc44a37e
changed types like the return type of `Cache.get` to be non-null instead
of unspecified.

As discussed in https://github.com/ben-manes/caffeine/issues/594, the
types really "should" be nullable, but that is inconvenient for most
users.

When Caffeine was using Checker Framework annotations, it could use
`@PolyNull` to give Checker Framework users the best of both worlds and
give users of other tools (notably Kotlin) a convenient (if not fully
strict) experience, since those tools viewed the types as having
unspecified nullness (in Kotlin terms, "platform types").

Caffeine could still use `@PolyNull` today if you want to restore the
Checker Framework dependency. (And maybe someday JSpecify will include
its own version, as discussed in
https://github.com/jspecify/jspecify/issues/79.)

But the important thing is that Caffeine _not_ affirmatively make the
type be non-null. And that's what `@NullMarked` does. That would cause
trouble for a number of Kotlin users in Google's codebase.

So, to undo that, we can use `@NullUnmarked` (recognized by Kotlin as of
2.0.20) on the methods where we want to keep types unspecified. Then we
can optionally use `@NonNull` annotations on the _other_ types in the
API in order to still make those types non-null.

Kotlin actually _still_ doesn't quite get the types right, thanks to
https://youtrack.jetbrains.com/issue/KT-73658. But this PR at least
annotates correctly (I hope :)), which may help other tools even today
and which should help Kotlin eventually.

If enough people run into the Kotlin bug in practice, then we could
consider backing out even more nullness information: If we were to
remove `@NullMarked` from `Cache`+`AsyncCache` and then potentially
sprinkle in some `@NonNull` annotations to compensate, then we could
still preserve at least some of the new information in those classes.
And of course other classes can remain `@NullMarked`.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
@@ -198,7 +198,8 @@ public interface AsyncCache<K, V> {
    *
    * @return a thread-safe view of this cache supporting all of the optional {@link Map} operations
    */
-  ConcurrentMap<K, CompletableFuture<V>> asMap();
+  @NullUnmarked
+  @NonNull ConcurrentMap<@NonNull K, @NonNull CompletableFuture<V>> asMap();
 
   /**
    * Returns a view of the entries stored in this cache as a synchronous {@link Cache}. A mapping is
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
@@ -23,7 +23,9 @@ import java.util.concurrent.Executor;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
+import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 
 /**
@@ -47,7 +49,7 @@ public interface AsyncCache<K, V> {
    *
    * @param key the key whose associated value is to be returned
    * @return the future value to which the specified key is mapped, or {@code null} if this cache
-   *         does not contain a mapping for the key
+   *     does not contain a mapping for the key
    * @throws NullPointerException if the specified key is null
    */
   @Nullable
@@ -71,7 +73,9 @@ public interface AsyncCache<K, V> {
    * @return the current (existing or computed) future value associated with the specified key
    * @throws NullPointerException if the specified key or mappingFunction is null
    */
-  CompletableFuture<V> get(K key, Function<? super K, ? extends V> mappingFunction);
+  @NullUnmarked
+  @NonNull CompletableFuture<V> get(
+      @NonNull K key, @NonNull Function<? super @NonNull K, ? extends V> mappingFunction);
 
   /**
    * Returns the future associated with the {@code key} in this cache, obtaining that value from
@@ -89,15 +93,22 @@ public interface AsyncCache<K, V> {
    *
    * @param key the key with which the specified value is to be associated
    * @param mappingFunction the function to asynchronously compute a value, optionally using the
-   *        given executor
+   *     given executor
    * @return the current (existing or computed) future value associated with the specified key
-   * @throws NullPointerException if the specified key or mappingFunction is null, or if the
-   *         future returned by the mappingFunction is null
-   * @throws RuntimeException or Error if the mappingFunction does when constructing the future,
-   *         in which case the mapping is left unestablished
+   * @throws NullPointerException if the specified key or mappingFunction is null, or if the future
+   *     returned by the mappingFunction is null
+   * @throws RuntimeException or Error if the mappingFunction does when constructing the future, in
+   *     which case the mapping is left unestablished
    */
-  CompletableFuture<V> get(K key, BiFunction<? super K, ? super Executor,
-      ? extends CompletableFuture<? extends V>> mappingFunction);
+  @NullUnmarked
+  @NonNull CompletableFuture<V> get(
+      @NonNull K key,
+      @NonNull
+          BiFunction<
+              ? super @NonNull K,
+              ? super @NonNull Executor,
+              ? extends @NonNull CompletableFuture<? extends V>>
+          mappingFunction);
 
   /**
    * Returns the future of a map of the values associated with the {@code keys}, creating or
@@ -118,11 +129,11 @@ public interface AsyncCache<K, V> {
    * @param keys the keys whose associated values are to be returned
    * @param mappingFunction the function to asynchronously compute the values
    * @return a future containing an unmodifiable mapping of keys to values for the specified keys in
-   *         this cache
+   *     this cache
    * @throws NullPointerException if the specified collection is null or contains a null element, or
-   *         if the future returned by the mappingFunction is null
+   *     if the future returned by the mappingFunction is null
    * @throws RuntimeException or Error if the mappingFunction does so, in which case the mapping is
-   *         left unestablished
+   *     left unestablished
    */
   CompletableFuture<Map<K, V>> getAll(Iterable<? extends K> keys,
       Function<? super Set<? extends K>, ? extends Map<? extends K, ? extends V>> mappingFunction);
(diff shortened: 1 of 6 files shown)


caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
```diff
@@ -170,7 +181,7 @@ CompletableFuture<Map<K, V>> getAll(Iterable<? extends K> keys,
    * @param valueFuture the value to be associated with the specified key
    * @throws NullPointerException if the specified key or value is null
    */
-  void put(K key, CompletableFuture<? extends V> valueFuture);
+  void put(K key, CompletableFuture<? extends @Nullable V> valueFuture);
```
cpovirk, line 184: This is basically unrelated. I just noticed that it's another case in which null might actually be valid, and it's a case in which there's no _return type_ that we might choose to _also_ make nullable.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
```diff
@@ -71,7 +73,9 @@ public interface AsyncCache<K, V> {
    * @return the current (existing or computed) future value associated with the specified key
    * @throws NullPointerException if the specified key or mappingFunction is null
    */
-  CompletableFuture<V> get(K key, Function<? super K, ? extends V> mappingFunction);
+  @NullUnmarked
+  @NonNull CompletableFuture<V> get(
+      @NonNull K key, @NonNull Function<? super @NonNull K, ? extends V> mappingFunction);
```
cpovirk, line 78: Actually, maybe I should include `@Nullable` for the output type of the functions? That would work around the Kotlin bug that I mentioned, and it would still be accurate. All that we really care about is keeping the return type convenient. I may give that a try shortly....

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
```diff
@@ -170,7 +181,7 @@ CompletableFuture<Map<K, V>> getAll(Iterable<? extends K> keys,
    * @param valueFuture the value to be associated with the specified key
    * @throws NullPointerException if the specified key or value is null
    */
-  void put(K key, CompletableFuture<? extends V> valueFuture);
+  void put(K key, CompletableFuture<? extends @Nullable V> valueFuture);
 
   /**
    * Returns a view of the entries stored in this cache as a thread-safe map. Modifications made to
```
cpovirk, line 187: Sorry, I also meant to ask whether `@Nullable V` would make sense for the type argument inside `asMap()`. That, too, though, is unrelated to the main change.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
ben-manes: I agree, if we need to use `@NullUnmarked` then we should be fully explicit because it is confusing to have defaults differ between methods.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
ben-manes: I think so because we always return the user's future and not any of our own dependent actions. There are a variety of technical reasons for this wrt correctness, but it means if their future resolves to null or an error state then it is observable. The cache will discard it by a dependent action.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
cpovirk: Resolving based on the thumbs-up emoji.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
cpovirk: Thanks. I'm guessing this is another case in which `@Nullable` would be inconvenient, so I've changed this to also be null-unmarked with per-type annotations.

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
cpovirk: OK, I've added `@Nullable` to all the function output types. As expected, this appears to be helping with the Kotlin issues (though my testing is still underway).

ben-manes (approved):

ben-manes:
Feel welcome to merge when you are comfortable with the changes.

cpovirk:
Thanks for the review, and thanks for giving the JSpecify annotations a spin in the first place!

It looks like I may end up wanting to touch `AsyncCacheLoader` and `CacheLoader` in similar ways. Let me know if you have any preference between adding those changes to this PR (and giving you another chance to review) and submitting a separate PR.

cpovirk:
Thanks. I will be waiting to hear back from our open-source-licensing people in any case. Hopefully the process will be relatively quick, since (I think you've said) your CLA is basically our CLA :)

ben-manes:
It is just the Apache Foundation's, which merely reiterates the license as a notice for an optional explicit agreement instead of an implicit one through the act of contributing. Its entirely optional if not signing is preferable.

> The CCLA is a backup document that the committer/ICLA signer may use to eliminate all of the ambiguity between conflicting laws, contracts, policies and job assignments. We've never required it. Many committers are confident of their individual representations under the ICLA, many other committers find it reassuring that their company has backed up their own ICLA with this umbrella document.

https://www.apache.org/licenses/cla-faq.html#cclas-not-required

cpovirk:
Thanks. If the licensing people take a while to get back to me, then I can merge without signing.

I've added the changes to the loader classes. I'll probably make sure that I have all our depot's tests passing before I merge anything here.

(I could believe that there are other types in Caffeine that could still benefit from annotations, especially more of those found in type arguments. I'm not currently planning to look proactively, but we'll see if any more come to my attention or if I go looking for a distraction :))

cpovirk:
I found a few more, not through build/test failures but just through further examination.

And I was doing the further examination because I am trying to get together an alternative approach: We could allow the various cache classes to have a nullable value type. That wouldn't mean that the caches could "contain" `null`, since they never cache that value. What it would mean is that the cache uses/supports a loader that can _return_ `null`, which in turn means that the computing get methods can also return `null`. That would allow us to give a specified return type to all the methods, with users able to declare either "a cache that might return null" or "a cache that will never return null." I have an attempt at that in a Piper CL that I'm testing against the depot. I'll see how it looks, and then I'll put together a separate PR so that you can see what you think.

In the meantime, I'm testing the latest version of this PR :)

ben-manes:
oh that sounds pretty amazing if it can work that way!

If helpful, the [examples](https://github.com/ben-manes/caffeine/tree/master/examples) section acts as integration tests using separate builds. This way I could avoid complexity in the main build, like Graal AOT set up, by isolating those tooling scenarios. Then I write it up under the guise of a tutorial, somewhat to help me remember how it works, and I have regression test that can run against the [locally built jar](https://github.com/ben-manes/caffeine/blob/2542ca8ca22fe4c30b62ee569ee71436fc44a37e/examples/graal-native/build.gradle.kts#L38-L44) in CI. That approach might be attractive to you if nullability is currently best verified in Kotlin so that I don't mangle that in the future.

cpovirk:
Great!

My new attempt at annotating seems to be working well: It requires fewer updates to users (not that I needed to make too many changes for your original change plus this PR, but less is nice), and it ensures that anyone who passes a null-returning loader will have to handle null during cache operations. It's passing all our tests.

Meanwhile, I did get this PR passing all our tests, as well. My plan is to merge this one (to at least immediately eliminate the non-null types that were my main worry) and then create another PR that shows my new attempt as a diff against this one.

(I even got approval for the CLA :))

cpovirk:
Sorry, I just noticed that the default is "Merge pull request" but that the Caffeine history looks linear, at least recently. It might have been better for me to either squash or rebase.

ben-manes:
No problem. I disabled the other options going forward so squash is required. There is no default, only the last option selected by the user.

PR #1806, opened 2024-12-05
Allow caches to have nullable type arguments for their value types.

cpovirk:
This does not mean that the cache will "contain" `null`, since `null` is
never cached. However, it does allow users to declare whether they want
"a cache whose loads might return `null`" (which thus can return `null`
from cache operations) or "a cache whose loads will never return `null`"
(which thus doesn't require users to handle `null` when they request
that the cache return or load a value).

This PR follows up on https://github.com/ben-manes/caffeine/pull/1805.

Ideally I will come back to write some Kotlin tests for this. For the
moment, I can only say that the results look good in my testing inside
Google.

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java
@@ -25,7 +25,6 @@ import java.util.function.Function;
 
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullMarked;
-import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 
 /**
@@ -41,7 +40,7 @@ import org.jspecify.annotations.Nullable;
  * @param <V> the type of mapped values
  */
 @NullMarked
-public interface AsyncCache<K, V> {
+public interface AsyncCache<K, V extends @Nullable Object> {
 
   /**
    * Returns the future associated with the {@code key} in this cache, or {@code null} if there is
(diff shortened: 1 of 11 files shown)


ben-manes (approved):
Nice!

When adopting jspecify, I had forgotten that  tests disabled NullAway. It probably makes sense for me to revisit that as a safety net for reviewing nullability changes.

https://github.com/ben-manes/caffeine/blob/047021c523d9511f2747dc02f9bdeda35dd17269/gradle/plugins/src/main/kotlin/quality/errorprone.caffeine.gradle.kts#L65-L68

cpovirk:
I'll come back to have a look at [the JCache error](https://github.com/ben-manes/caffeine/actions/runs/12183031590/job/33983468339?pr=1806). I'm hoping that I'll just need to twiddle some annotations, but it may be that we are running up against the frontier of what NullAway can do for generics so far.

ben-manes:
I'm unsure why it only failed on one compile and not elsewhere, where it somehow decided it was eligible to retrieve it from the build cache. The build uses the common env variable `CI=true` to enable `-Werror` so that local dev builds can observe all issues rather than fail eagerly at the first. You can use `--rerun` or `--no-build-cache` if that interferes at all, and swap JDK versions using `JAVA_VERSION=?` where it defaults to `11`. Locally it failed right away.

cpovirk:
Thanks. I thought I had a problem last week, and then it passed a local build today, and so I pushed it to start the rest of the testing, since I figured it couldn't be that easy :) In parallel, I kicked off a build with `CI=true`, and that was enough to get it to fail locally again. So the latest commit may or may not be of value. I'll see what I can figure out....

ben-manes:
SonarQube also has [false positives](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=com.github.ben-manes.caffeine%3Acaffeine&open=AZOx5AKR8_r8rK9wQvm4) on JSpecify, but those don't break the build so I am suppressing them remotely. It thinks that `Cache.asMap()` is nullable which is not the case.

Screenshot 2024-12-10 at 1 05 24 PM

PR #1835, opened 2025-02-26
Remove some unneeded suppressions

msridhar:
Found by locally compiling with a build of Error Prone with https://github.com/google/error-prone/pull/4828 applied and a NullAway branch that supports the unused suppression detection.  (Support for JdkObsolete was built into the Error Prone PR)


ben-manes (approved):
Awesome!

PR #1842, opened 2025-03-26
Update to NullAway 0.12.6

msridhar:
Fix or suppress newly-reported issues


ben-manes:
sgtm, thank you!

caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsyncAsMapTest.java
```diff
@@ -2594,7 +2595,7 @@ public void entrySet_remove_null(AsyncCache<Int, Int> cache, CacheContext contex
   @Test(dataProvider = "caches")
   public void entrySet_remove_nullKey(AsyncCache<Int, Int> cache, CacheContext context) {
     var future = Iterables.getFirst(cache.asMap().values(), context.absentValue().toFuture());
-    assertThat(cache.asMap().entrySet().remove(Maps.immutableEntry(null, future))).isFalse();
+    assertThat(cache.asMap().entrySet().remove(Maps.<@Nullable Int, CompletableFuture<Int>>immutableEntry(null, future))).isFalse();
```
msridhar, line 2598: These explicit type arguments are needed due to limitation in NullAway's inference.  They kind of clutter the code, so we could just suppress instead

caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsMapTest.java
```diff
@@ -2792,7 +2792,7 @@ public void entrySet_remove_null(Map<Int, Int> map, CacheContext context) {
 
   @CacheSpec
   @CheckNoStats
-  @SuppressWarnings("MapEntry")
+  @SuppressWarnings({"MapEntry","NullAway"})
```
msridhar, line 2795: Here I suppressed rather than adding explicit type arguments (compare with `AsyncAsMapTest` fixes).

msridhar:
@ben-manes this passes NullAway 0.12.6 now, except for a couple of `Map.get` calls where I wasn't sure if they could actually return `null` so I didn't suppress; was hoping you could look at those.  All other suppressions are due to remaining limitations in NullAway's inference of type arguments for generic method calls.

Generally speaking probably good if you look over all the changes since I don't know this code well 🙂

caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsyncAsMapTest.java
ben-manes: either is fine with me

msridhar: Ok, let me change to suppress.  When we improve NullAway's inference we then should be able to detect the unnecessary suppression and remove it

ben-manes:
thanks! I'll fix them. I believe in these cases I just add `requireNonNull` since it is implicit test state, so invalid if null (and that caught one or two mistakes!), and satisfies the checker.

PR #1851, opened 2025-04-15
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling)

The misspellings have been reported at https://github.com/jsoref/caffeine/actions/runs/14469708766#summary-40580264051

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/caffeine/actions/runs/14469693878#summary-40580214489

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
@@ -401,7 +401,7 @@ abstract class AbstractLinkedDeque<E> extends AbstractCollection<E> implements L
 
     @Override
     public boolean hasNext() {
-      checkForComodification();
+      checkForConcurrentModification();
       return (cursor != null);
     }
 
@@ -429,7 +429,7 @@ abstract class AbstractLinkedDeque<E> extends AbstractCollection<E> implements L
       if (previous == null) {
         throw new IllegalStateException();
       }
-      checkForComodification();
+      checkForConcurrentModification();
 
       AbstractLinkedDeque.this.remove(previous);
       expectedModCount = modCount;
@@ -440,7 +440,7 @@ abstract class AbstractLinkedDeque<E> extends AbstractCollection<E> implements L
      * If the expected modCount value that the iterator believes that the backing deque should have
      * is violated then the iterator has detected concurrent modification.
      */
-    void checkForComodification() {
+    void checkForConcurrentModification() {
       if (modCount != expectedModCount) {
         throw new ConcurrentModificationException();
       }
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
@@ -390,7 +390,7 @@ abstract class AbstractLinkedDeque<E> extends AbstractCollection<E> implements L
     int expectedModCount;
 
     /**
-     * Creates an iterator that can can traverse the deque.
+     * Creates an iterator that can traverse the deque.
      *
      * @param start the initial element to begin traversal from
      */
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/ExpireAfterVarTest.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/ExpireAfterVarTest.java
@@ -563,7 +563,7 @@ public final class ExpireAfterVarTest {
 
   @Test(dataProvider = "caches")
   @CacheSpec(population = Population.FULL, expiry = CacheExpiry.MOCKITO)
-  public void replaceConditionally_expiryFails_asybc(
+  public void replaceConditionally_expiryFails_async(
       AsyncCache<Int, Int> cache, CacheContext context) {
     when(context.expiry().expireAfterCreate(any(), any(), anyLong()))
         .thenThrow(ExpirationException.class);
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/testing/CacheGenerator.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/testing/CacheGenerator.java
@@ -217,7 +217,7 @@ public final class CacheGenerator {
   @SuppressWarnings("unchecked")
   private static void populate(CacheContext context, Cache<Int, Int> cache) {
     if (context.population.size() == 0) {
-      // timeWhel clock initialization
+      // timerWheel clock initialization
       cache.cleanUp();
       return;
     }
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/TraceReader.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/TraceReader.java
@@ -46,7 +46,7 @@ public interface TraceReader {
   @MustBeClosed
   Stream<AccessEvent> events();
 
-  /** A trace reader that that does not contain external event metadata. */
+  /** A trace reader that does not contain external event metadata. */
   @FunctionalInterface
   interface KeyOnlyTraceReader extends TraceReader {
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -363,7 +363,7 @@ jobs:
             echo -e '\n'
             echo '#### [Get/Put](https://jmh.morethan.io/?gists=b231cf57cf8e144e2247716e777edcf3)'
             echo -n 'A benchmark that evaluates the read/write performance of a cache. The cache '
-            echo -n 'is pre-populated for a 100% hit rate and a Zipf distribution of keys is used '
+            echo -n 'is prepopulated for a 100% hit rate and a Zipf distribution of keys is used '
             echo -n 'to mimic application usage patterns.'
             echo -e '\n'
             echo '#### [Put/Remove](https://jmh.morethan.io/?gists=1359c399cb717e81976ee635c513340b)'
--- a/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/EvictionBenchmark.java
+++ b/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/EvictionBenchmark.java
@@ -24,7 +24,7 @@ import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
 
 /**
- * A benchmark that evaluates the eviction performance of a cache. The cache is pre-populated for
+ * A benchmark that evaluates the eviction performance of a cache. The cache is prepopulated for
  * a 100% eviction rate to mimic worst case behavior.
  * <p>
  * {@snippet :
--- a/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/GetPutBenchmark.java
+++ b/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/GetPutBenchmark.java
@@ -32,7 +32,7 @@ import site.ycsb.generator.NumberGenerator;
 import site.ycsb.generator.ScrambledZipfianGenerator;
 
 /**
- * A benchmark that evaluates the read/write performance of a cache. The cache is pre-populated for
+ * A benchmark that evaluates the read/write performance of a cache. The cache is prepopulated for
  * a 100% hit rate and a Zipf distribution of keys is used to mimic application usage patterns.
  * <p>
  * {@snippet :
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java
@@ -413,7 +413,7 @@ public final class Caffeine<K, V> {
         "maximum size was already set to %s", this.maximumSize);
     requireState(this.maximumWeight == UNSET_INT,
         "maximum weight was already set to %s", this.maximumWeight);
-    requireState(this.weigher == null, "maximum size can not be combined with weigher");
+    requireState(this.weigher == null, "maximum size cannot be combined with weigher");
     requireArgument(maximumSize >= 0, "maximum size must not be negative");
     this.maximumSize = maximumSize;
     return this;
@@ -491,7 +491,7 @@ public final class Caffeine<K, V> {
     requireNonNull(weigher);
     requireState(this.weigher == null, "weigher was already set to %s", this.weigher);
     requireState(!strictParsing || this.maximumSize == UNSET_INT,
-        "weigher can not be combined with maximum size");
+        "weigher cannot be combined with maximum size");
 
     @SuppressWarnings("unchecked")
     var self = (Caffeine<K1, V1>) this;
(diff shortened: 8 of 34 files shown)


caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/NodeFactoryGenerator.java
```diff
@@ -206,7 +206,7 @@ private static Set<List<Object>> combinations() {
   private static String encode(String className) {
     return Feature.makeEnumName(className)
         .replaceFirst("STRONG_KEYS", "P") // puissant
-        .replaceFirst("WEAK_KEYS", "F") // faible
+        .replaceFirst("WEAK_KEYS", "F") // failable
```
jsoref, line 209: Note that the normal word is `fallible`, but I'm not really sure what this project is doing and it was already using `failible` ...

caffeine/src/main/java/com/github/benmanes/caffeine/cache/AbstractLinkedDeque.java
```diff
@@ -440,7 +440,7 @@ public void remove() {
      * If the expected modCount value that the iterator believes that the backing deque should have
      * is violated then the iterator has detected concurrent modification.
      */
-    void checkForComodification() {
+    void checkForConcurrentModification() {
```
jsoref, line 443: This word is too close to Commodification which is something else, and as this is not a public function, it seemed reasonable to spell it out (java isn't particularly shy about long method names...)

caffeine/src/test/java/com/github/benmanes/caffeine/cache/testing/CacheGenerator.java
```diff
@@ -217,7 +217,7 @@ private static <K, V> Cache<K, V> newCache(CacheContext context) {
   @SuppressWarnings("unchecked")
   private static void populate(CacheContext context, Cache<Int, Int> cache) {
     if (context.population.size() == 0) {
-      // timeWhel clock initialization
+      // timeWheel clock initialization
```
jsoref, line 220: ??

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/TraceReader.java
```diff
@@ -46,7 +46,7 @@ public interface TraceReader {
   @MustBeClosed
   Stream<AccessEvent> events();
 
-  /** A trace reader that that does not contain external event metadata. */
+  /** A trace reader that does not contain external event metadata. */
```
jsoref, line 49: Sometimes `that that` is correct. This isn't one of those times...

caffeine/src/test/java/com/github/benmanes/caffeine/cache/ExpireAfterVarTest.java
```diff
@@ -563,7 +563,7 @@
 
   @Test(dataProvider = "caches")
   @CacheSpec(population = Population.FULL, expiry = CacheExpiry.MOCKITO)
-  public void replaceConditionally_expiryFails_asybc(
+  public void replaceConditionally_expiryFails_async(
```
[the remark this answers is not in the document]
jsoref: It sounds like Codacy wants this:
```suggestion
  public void replaceConditionallyExpiryFailsAsync(
```

caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/NodeFactoryGenerator.java
ben-manes: The contributor who refactored this code (https://github.com/ben-manes/caffeine/pull/200) used french words to avoid name collisions. This context is for a weak reference which allows the garbage collector to discard it when there are no strong references.

> "Faible" is a French word meaning weak, and "failable" is an English word meaning liable to make mistakes or fail. The key difference is that "faible" describes weakness, while "failable" describes the potential to fail.

caffeine/src/test/java/com/github/benmanes/caffeine/cache/ExpireAfterVarTest.java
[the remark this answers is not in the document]
ben-manes: ❌ 

Codacy blindly runs a bunch of 3rd party tools unconfigured and I haven't spent the time trying to get it to conform to our settings (in this case PMD v6, whereas we run v7 directly and configure it). We follow the Google Java Style Guide where the [naming convention](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names) is,

> Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), for example transferMoney_deductsFromSource. There is no One Correct Way to name test methods.

caffeine/src/test/java/com/github/benmanes/caffeine/cache/testing/CacheGenerator.java
ben-manes: meant to be `timerWheel`. This is a data structure (see TimerWheel.java) also referred to as a timing wheel. It is an O(1) priority queue based on hashing.

ben-manes (approved):
Thank you! Please revert the ❌ and I will merge. 🙂 

Some of these are from imported code (test suite, research references) so I am not against changing, just that I did the minimal to incorporate them.

Many others are my goofs so this is really great. I have a [spelling action](https://github.com/ben-manes/caffeine/blob/master/.github/workflows/spelling.yml) that run misspell and typos which missed all of these. Looks like yours should be adding to that list.

caffeine/src/test/java/com/github/benmanes/caffeine/cache/ExpireAfterVarTest.java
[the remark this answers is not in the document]
jsoref: Sounds reasonable. I personally wouldn't have wanted to make the change that Codacy is implying. If I return here, I might spend the time to see if I can beg it not to make this request because it definitely seems annoying.

jsoref (commented):
I'll drop the ❌ items and fix `timerWheel`.

Yes, I saw you had a `spelling` workflow and was amused that it didn't catch any of these items. I didn't spend time looking into which thing it was (but it did seem likely that it was a codespell/misspell thing which is basically a fixed set of rules for typos as opposed to something that works based on dictionaries).

If you're interested in check-spelling, you'd want to use a released version as opposed to prerelease (more or less that just means changing the uses: lines to reference the latest release by sha).

ben-manes:
perfect, thanks for the help!

I am using [typos](https://github.com/crate-ci/typos) and [misspell](https://github.com/golangci/misspell). Both are based on the [correction approach](https://github.com/crate-ci/typos/blob/master/docs/design.md#trade-offs) rather than dictionaries. I'll look into integrating your github action integrated this evening.

caffeine/src/javaPoet/java/com/github/benmanes/caffeine/cache/NodeFactoryGenerator.java
ben-manes: sgtm, I'd probably use `persistent` since this is about strong-weak-soft references for whether the instance can disappear underneath you.

ben-manes:
I [set up](https://github.com/ben-manes/caffeine/actions/runs/14483789725/workflow#L29) your action using your commit. A little feedback,
1. I use Harden Runner to restrict network access to avoid compromised actions. Yours requires sudo, can that be changed to allow it to be disabled? 
2. I use actions/checkout explicitly to disable `persist-credentials` which leaks the token. Your action has an option to checkout, but it was unclear if that used this protection. It's a good practice and likely the repo owner should handle the checkout because of that.
3. For `use_sarif` it is nice to have it on branches, so I restricted with `github.event.pull_request.head.repo.fork` instead
4. It seemed to need the `.github/actions/spelling` files on master so I couldn't iterate on a branch.
5. I had to add `.gitleaksignore` afterwards due to the `block-delimiters.list` and was a false positive.
6. Codacy had useless complaints from markdownlint (ignored) 😆

ben-manes:
oh yeah, and for some reason you have an insecure http call to azure.archive.ubuntu.com. Can that be https?

jsoref:
1. I've played w/ [Harden Runner](https://github.com/step-security/harden-runner) occasionally... I'm using `sudo` to install software w/ `apt-get`. I could replace it with stuff to do local installs of software, but it'd require some amount of work. I might try today. 📝
2. w.r.t. `persist-credentials`, I could change `checkout` to suppress it, but as there's a mode where it actually needs persist-credentials, making it magical seemed too hard for people to reason through, so my initial inclination was for them to use actions/checkout themselves if they're concerned.
  I'll either add a note indicating that it doesn't disable persist-credentials or add the code to let it try to do the right thing (https://github.com/check-spelling-sandbox/check-spelling/commit/315e15648a92867367fc36c3306c10bf262b399a) 📝.
3. For `use_sarif`, I think the logic I'm using, while convoluted, should do what you want:
    ```yaml
    use_sarif: >-
      (!github.event.pull_request # this covers pushes to any branches in the active repository -- or really anything that isn't a pull_request, e.g., in theory a `workflow_dispatch` although you'd need to use `input.event_aliases` to tell check-spelling what to do if you used that event)
       ||
       (github.event.pull_request.head.repo.full_name == github.repository) # this covers pull requests within the same repository
      )
      && 1 }}
    ```
    If there's better documentation that would help you or another reader understand the logic, I'm more than happy to incorporate it. 📝
4. For `.github/actions/spelling` files on a branch, I don't suppose you could point to the run where this happened?
  That's probably poor messaging on my part.
  I do all of my development work for PRs like this on a branch, e.g.: https://github.com/check-spelling-sandbox/caffeine/actions/workflows/check-spelling.yml?query=branch%3Aspell-check-with-spelling
5. For `.gitleaksignore`, and `block-delimiters.list`, you should probably just delete both -- `block-delimiters.list` hasn't shipped in a release because I'm still iterating on the feature. I'm still gathering implementation experience/trying to think of a better design for the file. I don't know how check-spelling could help inform users that random files in config aren't supported by a given version.
  I suppose I could have a simple thing that marks all the files it is consuming and warns about anything else, that should at least give people a hint when they try to use a future feature in an older version 📝. I might look into that today (at best that'd let v0.0.25 complain about this file, I obviously can't retroactively improve v0.0.24).
6. Re: complaints from `markdownlint`, I'm not opposed to changing my markdown to make people happier (I have tried to change it a bit in response to some nudges from prettier). I'm not a huge fan of any of the linters and I don't really like some of the required styles (especially the ones that require full whitespace indentation for table columns because that results in a lot of jitter).
7.  I was a bit puzzled by `http` and `azure.archive.ubuntu.com`. The who was obviously `apt-get`, and it was obviously to install something, but it wasn't particularly obvious as I definitely didn't ask for `http:`.
    With a little debugging in https://github.com/check-spelling-sandbox/crispy-train-expert-chainsaw/actions/runs/14492227398, it turns out that `/etc/apt/sources.list.d/ubuntu.sources` refers to `/etc/apt/apt-mirrors.txt` which in turn references http://azure.archive.ubuntu.com/ubuntu/ and other `http:` services.
    As I'm _already_ rewriting some of those files, I could _probably_ try to replace certain well known urls.
    It's a bit problematic as I can't easily know if a thing definitely supports https without doing a lot of work. I also want to try to work on runners that aren't `linux/amd64`, so if someone were to run check-spelling on `linux/armhf` or `linux/arm64` (I run this locally on my mac using [nektos/act](https://github.com/nektos/act) talking to [lima-vm](https://github.com/lima-vm/lima)), I would still want things to work (I believe these two are officially supported by Ubuntu but if a future Ubuntu supported some other architecture, I would want check-spelling to work w/o having to ship an update for it).
    I'll definitely look into trying to do _something_ about this, as I agree, using `http` instead of `https` is silly.
    GitHub changed _some_ of the urls from `http` to `https`: https://github.com/actions/runner-images/blame/d3ec6220585779dd1e2a966ed468cb00b45de5cc/images/ubuntu/scripts/build/configure-apt-sources.sh#L11-L13
    The answer for "Why is GitHub's runner not using https for azure.archive.ubuntu.com?" is "because the server doesn't seem to support it 🤯". Give https://azure.archive.ubuntu.com/ a try yourself.
    
    That said, I have found `apt-get` to be painful & slow, so I might experiment with a mode that downloads programs directly instead...

ben-manes:
1. not a big deal, more an fyi if an easy fix
2. sgtm. I'm not actually concerned
3. oh I probably confused myself 🙂 
4. It might have been a misunderstanding because I had to update the exclusion list. The [error logs](https://github.com/ben-manes/caffeine/actions/runs/14483520607/job/40624824973) ended with the adoption information which I mistook as being about not being on the main branch yet.
5. oh its fine, just an fyi that it can be picked up by leak detectors. not a problem at all
6. haha, it was just amusing given earlier discussion. I don't care and I don't make my markdown comply. It was just unhappy at line lengths over 80c, so really just the pedantic noise.
7. weird. again not a problem, just for your awareness.

I'm not actually super into the security stuff. I just want to (a) avoid cve spam from resume padding, (b) shamed on HN because I goofed, (c) learn so I can apply the more useful techniques at work. Take all of this as informational and not things you need to act on, but can if of interest.

jsoref:
* Point 4, I clearly need to adjust that text. It's https://github.com/check-spelling/spell-check-this/blob/prerelease/.github/actions/spelling/advice.md
  There is an `if` in that text which was supposed to result in you ignoring the message, but it clearly didn't work. I'm open to ideas (I don't have any right now, but I will definitely stew on the problem some more and hopefully find improved language).
  The idea is to catch a weird corner case where someone has merged the advice into the default branch and someone else had an open PR to that same branch from before that merge, and they add another change which now triggers check-spelling with advice to make changes to the metadata which they don't yet have because their branch hasn't rebased/merged with the updated default branch.

```mermaid
---
config:
  gitGraph:
    rotateCommitLabel: false
---
gitGraph TB:
    commit
    branch feature
    checkout feature
    commit id: "has typos"
    merge main type:HIGHLIGHT id:"initial pull request (not merged)"
    checkout main
    branch spelling
    checkout spelling
    commit id: "add check-spelling"
    checkout main
    merge spelling id:"check-spelling is now active for PRs"
    checkout feature
    commit id: "more feature work"
    merge main type:REVERSE id:"confusing message?"

```

* Point 5, yeah, I was about to play with gitleaks (I started doing that and then got side-tracked by looking into step-security, and then mermaid-js). 
   Since it's all metadata that people can configure, they're free to take what they like/remove what they don't need/adjust things on either side. I should probably keep in mind that scanners are likely to think about that kind of content and may need a way for someone to annotate content to say "don't mind this line" / "don't mind the next line" etc. As I'm still gathering requirements for a fancier implementation, I'll try to keep this in mind (added to https://github.com/check-spelling/check-spelling/wiki/Feature:-Block-Ignore#concerns).

* Point 7 ended up being an interesting.
   I think I actually looked into it ages ago and at that point determined azure's mirror didn't serve https and promptly forgot about it.
   Between when I first started investigating to reply to you and now it turns out that github merged a thing that actually broke their support of mirrors. So there was actually a time-point where you could have blocked the azure :80 mirror and then had apt be happy w/ a public https mirror, but as of right now you can't (I made a pull request).
   In the PR that tried to add more `https` stuff and broke the mirror support, someone actually wondered about the lack of https support, so I've tried to suggest that maybe they should ask azure to support it since, maybe, they're GitHub affiliated (I'm not, I just touch way too many edges and am willing to look down a couple of layers).
   
   The fragility of all of this definitely makes me _less_ inclined to go for a mechanism that involves apt/apt-get. I may see how painful it is to use a mechanism that only involves public github services.

PR #1855, opened 2025-04-30
Attributing Copyrights to contributors in Apache license.

maddestructor:
The current copyright license wasn't attributed properly making it invalid. Updating that since we are using your project internally and while making proper attributions on our side, we noticed it. 

We therefore wanted to contribute back proper licensing for you in the process. You can switch it to your own username instead of caffeine contributors if you'd like.

This pull request updates the copyright information in the `LICENSE` file to reflect the correct year and copyright owner.

* [`LICENSE`](diffhunk://#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7L190-R190): Updated the copyright notice from a placeholder (`[yyyy] [name of copyright owner]`) to "2015 Caffeine Contributors" for accuracy.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2015 Caffeine Contributors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


ben-manes:
Can you provide a stronger justification for this change?

> The current copyright license wasn't attributed properly making it invalid. 

That is untrue. The license states `END OF TERMS AND CONDITIONS` and the change is in the `APPENDIX`, which is not part of the license's legality. It provides only an example format, but each source code file is attributed. The description clearly states the expectation of replace the brackets with the identifying information.

While I can understand it may be confusing, modifying the license even in benign ways can cause extra frustration.

maddestructor:
I am not willing to argue on this, just wanted to help.

ben-manes:
Its Apache License, so if you are contributing then you can assign copyright to whomever you prefer. That could be to the project, to yourself, it doesn't quite matter since you are allowed to retain it as part of the license. Instead they require the "Grant of Copyright License" to allow usage. The copyright owner shouldn't impact the license for this project or any users, and I am personally a fan of giving people credit for their hard work so a small author tag, copyright author, or whatever is good to keep.

Issue #1883, opened 2025-07-01
Spliterators for synchronous view of async cache reports incorrect characteristics


pkoenig10:
Spliterators for the synchronous view of an async cache reports that it is `Spliterator.SIZED`. `Spliterator.estimatedSize()` returns the size of the backing map, but the spliterator filters our values that are not ready. This violates the contract of `Spliterator.SIZED`, causing runtime failures when used with methods like `Stream.toList()`.

```java
AsyncCache<String, String> cache = Caffeine.newBuilder().buildAsync();

cache.put("key", new CompletableFuture<>());

Map<String, String> map = cache.synchronous().asMap();

// This succeeds, as expected
assertThat(map.entrySet().iterator().hasNext()).isFalse();

// This fails, the actual value is Spliterator.DISTINCT | Spliterator.SIZED | Spliterator.SUBSIZED
assertThat(map.entrySet().spliterator().characteristics())
        .isEqualTo(Spliterator.DISTINCT | Spliterator.NONNULL | Spliterator.CONCURRENT);

// This throws java.lang.IllegalStateException: End size 0 is less than fixed size 1
assertThat(map.entrySet().stream().toList()).isEmpty();

// Map.keySet() and Map.values() behave the same
```

ben-manes:
It looks like this is an implicit default from `Set` which we need to override. In Caffeine's own Spliterator subtypes it doesn't use `Spliterator.SIZED`, but I missed that it was implicitly added here. 

```java
/**
 * Creates a {@code Spliterator} over the elements in this set.
 *
 * <p>The {@code Spliterator} reports {@link Spliterator#DISTINCT}.
 * Implementations should document the reporting of additional
 * characteristic values.
 *
 * @implSpec
 * The default implementation creates a
 * <em><a href="Spliterator.html#binding">late-binding</a></em> spliterator
 * from the set's {@code Iterator}.  The spliterator inherits the
 * <em>fail-fast</em> properties of the set's iterator.
 * <p>
 * The created {@code Spliterator} additionally reports
 * {@link Spliterator#SIZED}.
 *
 * @implNote
 * The created {@code Spliterator} additionally reports
 * {@link Spliterator#SUBSIZED}.
 *
 * @return a {@code Spliterator} over the elements in this set
 * @since 1.8
 */
@Override
default Spliterator<E> spliterator() {
    return Spliterators.spliterator(this, Spliterator.DISTINCT);
}
```

vlsi:
[...]

PR #1884, opened 2025-07-01
Return correct characteristics from spliterators

pkoenig10:
Fixes https://github.com/ben-manes/caffeine/issues/1883

### DIFF
--- a/caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsMapTest.java
+++ b/caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsMapTest.java
@@ -55,6 +55,7 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
+import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsMapTest.java
```diff
@@ -3605,9 +3630,21 @@ public void entrySpliterator_trySplit(Map<Int, Int> map, CacheContext context) {
   @CacheSpec
   @CheckNoStats
   @Test(dataProvider = "caches")
-  public void entrySpliterator_estimateSize(Map<Int, Int> map, CacheContext context) {
+  public void entrySpliterator_estimateSize(Map<Int, Int> map, CacheContext context){
+        var spliterator = map.entrySet().spliterator();
+        assertThat(spliterator.estimateSize()).isEqualTo(context.initialSize());
+  }
```
ben-manes, line 3636: minor indent

caffeine/src/test/java/com/github/benmanes/caffeine/cache/AsMapTest.java
```diff
@@ -2309,6 +2310,18 @@ public void keySpliterator_estimateSize(Map<Int, Int> map, CacheContext context)
     assertThat(spliterator.estimateSize()).isEqualTo(context.initialSize());
   }
 
+  @CacheSpec
+  @CheckNoStats
+  @Test(dataProvider = "caches")
+  public void keySpliterator_characteristics(Map<Int, Int> map, CacheContext context) {
```
ben-manes, line 2316: We should probably add the same test cases to `AsyncAsMapTest` too. Unfortunately a lot of similarity between the tests but given different parameters as async/sync views (`Map<K, CompletableFuture<V>>` vs `Map<K, V>`).

ben-manes: yeah it does but their different views (`Cache.asMap()` & `AsyncCache.synchronous().asMap()` vs `AsyncCache.asMap()`). Thanks for reporting and jump on the fix!

pkoenig10: Ah that makes sense.

PR #1962, opened 2026-06-05
parse hex addresses as unsigned in address trace readers

sahvx655-wq:
The address readers feed each trace line through Long.parseLong(hex, 16) to build the cache key, but a hardware address is an unsigned 64-bit value. Any address with the top bit set, anything from 0x8000000000000000 upwards, is outside the signed long range, so parseLong throws a NumberFormatException and the whole event stream aborts partway through the trace. I hit this while pointing the penalties reader at a trace captured from a 64-bit host; the run died on the first high address rather than reporting a parse warning, which is why it took a moment to trace back to the radix conversion.

Switching both sites to Long.parseUnsignedLong keeps the full 64-bit address as the key, with the high half wrapping into a negative long exactly as the existing BigInteger(value, 16).longValue() readers (object-store, gradle) already do. The conversion belongs in the reader since that is the only place the textual address is turned into the long key the policies consume.


ben-manes:
Thanks! fwiw, the trace readers do tend to be a little brittle to whatever worked on their source data set rather than generally robust.

ben-manes (approved):

PR #1963, opened 2026-06-08
expand camelab requests as contiguous sector keys

sahvx655-wq:
The Camelab reader expands a multi-sector request by looping over the request size and emitting startAddress + i * 512 for each block, but the documented trace format records both the address and the request size in 512-byte sectors. A four-sector read at LBA 100 therefore covers sectors 100 through 103, yet the reader produced 100, 612, 1124 and 1636. I noticed the gaps while comparing Camelab behaviour against the other block readers; the 512-wide spacing meant overlapping requests could never collide, so every simulation over a Camelab trace under-reported its hit rate and inflated the working set.

The address is already a sector index, so the stride multiply was the bug: it mixed a sector base with a byte step. Every sibling block reader (arc, umass storage, snia systor/enterprise/k5cloud) treats its size field as a count and ranges over contiguous blocks, and Camelab was the only one multiplying the index by the sector size. Switching to LongStream.range over the sector span keeps the conversion in the reader where the textual fields become the long keys the policies consume, and drops the per-request array allocation along the way.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/camelab/CamelabTraceReader.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/camelab/CamelabTraceReader.java
@@ -27,7 +27,6 @@ import com.github.benmanes.caffeine.cache.simulator.parser.TraceReader.KeyOnlyTr
  * @author ben.manes@gmail.com (Ben Manes)
  */
 public final class CamelabTraceReader extends TextTraceReader implements KeyOnlyTraceReader {
-  static final long BLOCK_SIZE = 512;
 
   public CamelabTraceReader(String filePath) {
     super(filePath);
@@ -42,13 +41,9 @@ public final class CamelabTraceReader extends TextTraceReader implements KeyOnly
         return LongStream.empty();
       }
 
-      long startAddress = Long.parseLong(array[2]);
-      int requestSize = Integer.parseInt(array[3]);
-      long[] blocks = new long[requestSize];
-      for (int i = 0; i < requestSize; i++) {
-        blocks[i] = startAddress + (i * BLOCK_SIZE);
-      }
-      return LongStream.of(blocks);
+      long startBlock = Long.parseLong(array[2]);
+      int sequence = Integer.parseInt(array[3]);
+      return LongStream.range(startBlock, startBlock + sequence);
     });
   }
 }


ben-manes (commented):
reviewed against the [format spec](https://trace.camelab.org/2016/03/01/flash.html) and I think you're right that there are mistakes. However, when I try the MSRC traces from CameLab and SNIA (after disabling its weight characteristic) they don't match in either versions of this reader. I think its because the SNIA reader is caching the full entry whereas the CameLab just a block, so its not one-to-one. Trying to adjust so we can verify its consistent.

```console
  ➜  caffeine git:(v3.dev) ✗ ./gradlew simulator:run -q \
    -Dcaffeine.simulator.admission.0="Always" \
    -Dcaffeine.simulator.policies.0="linked.Lfu" \
    -Dcaffeine.simulator.files.paths.0="snia-cambridge:/Users/ben/Documents/traces/msr_cambridge/proj_0.csv.gz" \
    -Dcaffeine.simulator.maximum-size=1_000_000
  ╔════════════╤══════════╤═══════════╤═════════╤════════╤══════════╤═══════════╤═════════╤═════════╗
  ║ Policy     │ Hit Rate │ Miss Rate │ Hits    │ Misses │ Requests │ Evictions │ Steps   │ Time    ║
  ╠════════════╪══════════╪═══════════╪═════════╪════════╪══════════╪═══════════╪═════════╪═════════╣
  ║ linked.Lfu │  84.27 % │   15.73 % │ 444,406 │ 82,975 │  527,381 │         0 │ 527,381 │ 45.6 ms ║
  ╚════════════╧══════════╧═══════════╧═════════╧════════╧══════════╧═══════════╧═════════╧═════════╝%
  ➜  caffeine git:(v3.dev) ✗ ./gradlew simulator:run -q \
    -Dcaffeine.simulator.admission.0="Always" \
    -Dcaffeine.simulator.policies.0="linked.Lfu" \
    -Dcaffeine.simulator.files.paths.0="camelab:/Users/ben/Downloads/final-trace/proj_0.revised" \
    -Dcaffeine.simulator.maximum-size=1_000_000
  ╔════════════╤══════════╤═══════════╤════════════╤═══════════╤════════════╤═══════════╤════════════╤══════════╗
  ║ Policy     │ Hit Rate │ Miss Rate │ Hits       │ Misses    │ Requests   │ Evictions │ Steps      │ Time     ║
  ╠════════════╪══════════╪═══════════╪════════════╪═══════════╪════════════╪═══════════╪════════════╪══════════╣
  ║ linked.Lfu │  82.50 % │   17.50 % │ 15,522,318 │ 3,292,307 │ 18,814,625 │ 2,292,307 │ 18,814,625 │ 929.0 ms ║
  ╚════════════╧══════════╧═══════════╧════════════╧═══════════╧════════════╧═══════════╧════════════╧══════════╝
```

sahvx655-wq:
Went back through the readers to pin down where the divergence comes from, and your read looks right. The SNIA `CambridgeTraceReader` is the odd one out: it emits a single `AccessEvent.forKeyAndWeight(offset, size)` per line, so the key is the raw byte offset and the request size only ever shows up as the entry weight. With the weighted characteristic disabled that collapses to one unit-weight entry per request keyed on the offset, which is exactly the "full entry vs a block" you described. arc, umass storage and snia systor all do the opposite and map a request onto the contiguous range `[startBlock, startBlock + sequence)`, one key per block, which is what this PR brings CameLab in line with. That's why the request counts sit an order of magnitude apart and overlapping requests can never collide on the Cambridge side.

The other thing to watch is units. The MSR Cambridge CSV records offset and size in bytes, so to land in the same per-sector key space you'd need the systor treatment: `startBlock = offset / 512` and `sequence = ceil(size / 512)` before ranging. CameLab's `.revised` format already gives address and size in sectors, which is why this patch ranges over them directly with no divide. Once both readers expand into the same sector indices they should track each other, though I wouldn't bank on them being identical depending on whether CameLab kept the original LBAs when they republished the MSR set in the flash format. Happy to fold the Cambridge per-block change into this PR if that's the direction you're heading, otherwise I'll keep this one scoped to CameLab and leave that to your adjustment.

On the CodeQL flags: the raw `parseLong`/`parseInt` without a catch matches arc, storage and systor, so that's the existing convention in these readers rather than anything new in this diff.

ben-manes (approved):
> - Model: 
>     - Cambridge is object
>     - CameLab is page
>     - Fixable
> - Address space (the binding one): 
>     - SNIA is the original disk LBA — byte offsets, 82,975 distinct sectors, 16.24 GiB extent
>     - CameLab is TraceTracker's reconstructed flash address space — 52,611 distinct sectors, ~4 GiB extent
>     - Same workload, remapped addresses
> 
> So even if you convert Cambridge to a page model, the model would align but the footprints (82,975 vs 52,611) still differ, so the hit rates still diverge. No reader change fixes this — it would require inverting TraceTracker's remap, and that mapping isn't published. The divergence is the data, not the readers.
>
> The request-size distributions are identical — that's the verification you're after:
>
> | Size (sectors) | CameLab   | Cambridge |
> | -------------- | --------- | --------- |
> | 8              | 293,039   | 293,040   |
> | 128            | 63,844    | 63,844    |
> | 64             | 38,325    | 38,325    |
> | 16             | 23,862    | 23,862    |
> | 32             | 13,664    | 13,665    |
> | 24…104         | identical | identical |

PR #1964, opened 2026-06-09
read the object size as the weight in object store traces

sahvx655-wq:
The IBM ObjectStore records carry the object size as their own column, ahead of the offsets:

```
<time stamp> <request type> <object ID> <size of object> <beginning offset> <ending offset>
1232488 REST.GET.OBJECT 95d363d3fbdc0b03 1168 0 1167
```

`ObjectStoreTraceReader.events` was instead deriving the weight from the offsets as `end - start`, which came out a byte short of that size column on every GET, and where `start == end` (a single-byte read) collapsed to zero and got discarded by the `weight <= 0` guard. Per the [README](https://iotta.snia.org/traces/key-value/36305) the offsets are an inclusive range, so the documented object size is the cleaner source: read it directly from the size column instead. That drops the offset arithmetic, the guard, and the null-filtering tail, and lines this reader up with how `CambridgeTraceReader` reads its size field.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/keyvalue/ObjectStoreTraceReader.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/snia/keyvalue/ObjectStoreTraceReader.java
@@ -18,7 +18,6 @@ package com.github.benmanes.caffeine.cache.simulator.parser.snia.keyvalue;
 import static com.github.benmanes.caffeine.cache.simulator.policy.Policy.Characteristic.WEIGHTED;
 
 import java.math.BigInteger;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


ben-manes:
From the format description, it looks like we could read the weight directly.

>The format of the each trace record is <time stamp of request>, <request type>, <object ID>, <optional: size of object>, <optional: beginning offset>, <optional: ending offset>.  The timestamp is the number of milliseconds from the point where we began collecting the traces. For example:
>
>1219008 REST.PUT.OBJECT 8d4fcda3d675bac9 1056
1221974 REST.HEAD.OBJECT 39d177fb735ac5df 528
1232437 REST.HEAD.OBJECT 3b8255e0609a700d 1456
1232488 REST.GET.OBJECT 95d363d3fbdc0b03 1168 0 1167
1234545 REST.GET.OBJECT bfc07f9981aa6a5a 528 0 527
1256364 REST.HEAD.OBJECT c27efddbeef2b638 12752
1256491 REST.HEAD.OBJECT 13943e909692962f 9760
1256556 REST.GET.OBJECT 884ba9b0c6d1fe97 23872 0 23871
1256584 REST.HEAD.OBJECT d86b7bfefc63995d 12592

sahvx655-wq:
Agreed, that's cleaner. The size column sits right before the offsets (`REST.GET.OBJECT <id> <size> <start> <end>`), so I've switched `events()` to read the weight straight from `array[3]` instead of reconstructing it from the inclusive offsets. Pushed, and updated the title to match.

The whole guard falls out with it: the old `end - start` was a byte short of that size column on every GET, and the single-byte reads it collapsed to zero were being dropped by the `weight <= 0` check, so reading the documented size keeps those and lines this reader up with how `CambridgeTraceReader` pulls its own size field. I re-checked it against the README samples and the weights now match the recorded object sizes (1168, 528, and so on).

ben-manes (approved):

PR #1969, opened 2026-06-13
match the whole op field in the baleen trace reader

sahvx655-wq:
`BaleenTraceReader` picks the read requests out with `isRead(line[4].charAt(0))`, looking only at the first character of the op column. I went back to BCacheSim to pin down that field: it reads the op as `int(parts[4])` and maps it onto an `OpType` enum (`GET_TEMP=1`, `GET_PERM=2`, `PUT_TEMP=3`, `PUT_PERM=4`, `GET_NOT_INIT=5`, `PUT_NOT_INIT=6`, `UNKNOWN=100`), where the reads are exactly `{1, 2, 5}`. The leading-character test holds for the single-digit codes but trips on the one multi-digit value: `100` begins with `1`, so an `UNKNOWN` record is taken for a `GET_TEMP` and admitted as a cache read.

Comparing the whole op token keeps that case out. The distinction lives in the full value, so the filter has to look at all of it; once it is cut down to a leading character `100` cannot be told apart from `1`. Any Baleen trace carrying `UNKNOWN` operations would otherwise feed phantom requests into the run and skew the reported hit rate. The PUT codes are unaffected only because none of them starts with a read digit.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/baleen/BaleenTraceReader.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/baleen/BaleenTraceReader.java
@@ -40,7 +40,7 @@ public final class BaleenTraceReader extends TextTraceReader implements KeyOnlyT
     return lines()
         .dropWhile(line -> line.startsWith("#"))
         .map(line -> line.split(" ", 6))
-        .filter(line -> isRead(line[4].charAt(0)))
+        .filter(line -> isRead(line[4]))
         .flatMapToLong(line -> {
           long block = Long.parseLong(line[0]);
           long byteOffset = Long.parseLong(line[1]);
@@ -54,7 +54,7 @@ public final class BaleenTraceReader extends TextTraceReader implements KeyOnlyT
         });
   }
 
-  private static boolean isRead(char operation) {
-    return (operation == '1') || (operation == '2') || (operation == '5');
+  private static boolean isRead(String operation) {
+    return operation.equals("1") || operation.equals("2") || operation.equals("5");
   }
 }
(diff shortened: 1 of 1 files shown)


simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/baleen/BaleenTraceReader.java
```diff
@@ -54,7 +54,7 @@ public LongStream keys() {
         });
   }
 
-  private static boolean isRead(char operation) {
-    return (operation == '1') || (operation == '2') || (operation == '5');
+  private static boolean isRead(String operation) {
+    return operation.equals("1") || operation.equals("2") || operation.equals("5");
```
ben-manes: For the static analyzers, you can run `.github/scripts/analyze.sh` which produces this code quality recommendation. I suppose its not wrong that `isRead` could be replaced by a static ImmutableSet lookup, e.g. `.filter(READ_OPERATIONS::contains)`.

Screenshot 2026-06-13 at 11 44 25 AM

ben-manes (dismissed):

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/parser/baleen/BaleenTraceReader.java
sahvx655-wq: Done. Swapped it for a static READ_OPERATIONS set and dropped the isRead helper. The filter sees the split String[] rather than the op token on its own, so it reads as READ_OPERATIONS.contains(line[4]) instead of a bare method reference, but it is the same lookup you described.

I ran analyze.sh afterwards to be sure it was clean. ErrorProne nudged the field type from Set to ImmutableSet (PreferredInterfaceType/ImmutableMemberCollection), so I declared it as ImmutableSet and the build is warning-free now.

PR #1982, opened 2026-06-22
track the camp and gd-wheel weighted size as a long

sahvx655-wq:
Running the cost-aware policies over a byte-weighted trace with a maximum size above ~2 GiB, GD-Wheel quietly retained more than the configured limit and CAMP aborted in `finished()` with a `size != weightedSize` mismatch. Both track the resident weight in an `int size` field while `maximumSize` is a `long` and the per-entry weight comes straight off the trace, so once the cached weight passes `Integer.MAX_VALUE` the accumulator wraps negative. The `size > maximumSize` guards then stop firing, eviction never runs, and the cache grows past its bound; GD-Wheel reports an inflated hit rate and CAMP only catches it at shutdown when its weighted-size assertion trips.

Widen `size` to `long` in both, which is what the sibling weighted policies (Gdsf, Sieve, Linked, S3Fifo) already do. GD-Wheel summed the entry weights with `mapToInt` in `finished()`, so the consistency check overflowed in step with the field and hid the divergence; that sum is now `mapToLong` to match `GdsfPolicy`. Keeping the width right at the accumulator corrects every add and comparison without touching the eviction logic.

### DIFF
--- a/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
+++ b/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2026 Ben Manes. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.benmanes.caffeine.cache.simulator.policy.greedy_dual;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import com.github.benmanes.caffeine.cache.simulator.policy.AccessEvent;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * The cost-aware policies accumulate the cached weight, which the maximum size allows to exceed the
+ * signed int range. These checks fill a multi-gigabyte cache with byte-sized entries and confirm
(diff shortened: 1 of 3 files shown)


simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Test
+  void gdwheel() {
+    var policy = new GDWheelPolicy(config());
+    for (int key = 0; key < 4; key++) {
+      policy.record(AccessEvent.forKeyAndWeight(key, WEIGHT));
+    }
+    assertThat(policy.stats().evictionCount()).isGreaterThan(0L);
+  }
```
ben-manes: I think you can use an `@ParameterizedTest` with an `@MethodSource` so its one test against each policy.

simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
```diff
@@ -0,0 +1,62 @@
...
+import com.github.benmanes.caffeine.cache.simulator.policy.AccessEvent;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * The cost-aware policies accumulate the cached weight, which the maximum size allows to exceed the
+ * signed int range. These checks fill a multi-gigabyte cache with byte-sized entries and confirm
+ * the weighted size still bounds the cache so eviction fires.
+ *
+ * @author ben.manes@gmail.com (Ben Manes)
```
ben-manes: you can remove or set as yourself 🙂

ben-manes (dismissed):

simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
sahvx655-wq: Done. Collapsed both into a single @ParameterizedTest with a @MethodSource that supplies each policy constructor, so it runs once per policy now.

simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/greedy_dual/WeightedSizeOverflowTest.java
sahvx655-wq: Set it to myself, cheers.

PR #1983, opened 2026-06-23
warm up the clockprosimple hot set

sahvx655-wq:
Running the bundled LIRS traces past the policies, irr.ClockProSimple reports a 0% hit rate on loop at size 500 while irr.ClockPro and irr.ClockProPlus both score 48.86% on the same input, tying plain linked.Lru at zero. The fault is in onMiss: every first-access entry is linked into the cold list, so the hot list never fills. Since inTestPeriod returns true whenever sizeHot is zero, no cold entry is ever held long enough to be promoted, and the policy collapses to Clock behaviour over the whole key space and discards the working set it is meant to keep.

ClockPro and ClockProPlus seed the hot set during warm-up, giving newly accessed entries hot status until the resident size reaches maxSize minus minColdSize and only then inserting them cold in their test period. onMiss here was missing that branch even though the class doc states it works exactly like ClockPro. Adding the same warm-up restores loop to 48.86%, tracks ClockPro on multi1/multi3/cs within the small gap left by the consecutive-duplicate dedup the simple variant omits, and keeps the decision in onMiss beside the other first-touch handling where the sibling policies make it. The added test drives a loop larger than the cache and fails with zero hits on the current code.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProSimplePolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/ClockProSimplePolicy.java
@@ -174,10 +174,20 @@ public final class ClockProSimplePolicy implements KeyOnlyPolicy {
     policyStats.recordMiss();
     epoch++;
     var node = new Node(key, epoch);
-    node.status = Status.COLD;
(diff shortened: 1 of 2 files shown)


ben-manes:
@chanyoung this is your baby, any opinion?

PR #1984, opened 2026-06-23
skip consecutive duplicate accesses in lirs

sahvx655-wq:
The Lirs2 and ClockPro policies both collapse a rapid re-access to the same block (a correlated reference) to a single event at the top of record(), but irr.Lirs never picked up that guard. I noticed it while reconciling the irr family against Song Jiang's reference: Lirs2Policy's own comment points back at "the original LIRS conventions, which collapse such pairs to a single access", yet LirsPolicy counts "key, key" as a miss followed by a hit. The stack pruning and the LIR/HIR role swap are not idempotent, so a repeated key reorders stack S and can demote a different LIR block than a single access would, skewing the recency ordering and the reported hit rate on any trace with back-to-back duplicates.

Add the same lastKey guard the siblings use so a consecutive duplicate is ignored before any stack mutation. Keeping it in record() matches Lirs2 and ClockPro and the reference, where the dedup sits at the top of run_lirs; a non-consecutive reuse still counts as a genuine hit. The test covers both cases: "1, 1" records no hit, "1, 2, 1" records one.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/LirsPolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/LirsPolicy.java
@@ -77,6 +77,12 @@ public final class LirsPolicy implements KeyOnlyPolicy {
   int sizeHot;
   int residentSize;
 
+  // Skips consecutive duplicate accesses to a block. The stack pruning and LIR/HIR role swap are
+  // not idempotent, so counting "key, key" as two events diverges from Song Jiang's reference,
+  // which collapses such correlated references to a single access.
+  long lastKey = Long.MIN_VALUE;
+  boolean hasLastKey;
+
   public LirsPolicy(Config config) {
     var settings = new LirsSettings(config);
     this.maximumSize = Math.toIntExact(settings.maximumSize());
@@ -92,6 +98,12 @@ public final class LirsPolicy implements KeyOnlyPolicy {
 
   @Override
   public void record(long key) {
+    if (hasLastKey && (key == lastKey)) {
+      return;
+    }
+    lastKey = key;
+    hasLastKey = true;
+
     @Var @Nullable Node node = data.get(key);
     policyStats.recordOperation();
     if (node == null) {
--- a/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/LirsCorrelatedAccessTest.java
+++ b/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/LirsCorrelatedAccessTest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2026 Ben Manes. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.benmanes.caffeine.cache.simulator.policy.irr;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * LIRS treats a rapid re-access to the same block (a "correlated reference") as a single event, the
+ * convention its siblings Lirs2 and ClockPro already follow. These checks confirm a consecutive
+ * duplicate is collapsed while a genuine non-consecutive reuse is still counted.
+ *
+ * @author sahvx655@gmail.com (Sahana Bogar)
+ */
+final class LirsCorrelatedAccessTest {
+
+  @Test
+  void skipsConsecutiveDuplicate() {
+    var policy = new LirsPolicy(config());
+    policy.record(1);
+    policy.record(1);
+    assertThat(policy.stats().hitCount()).isEqualTo(0);
+  }
+
+  @Test
+  void countsNonConsecutiveReuse() {
+    var policy = new LirsPolicy(config());
+    policy.record(1);
+    policy.record(2);
+    policy.record(1);
+    assertThat(policy.stats().hitCount()).isEqualTo(1);
+  }
+
+  private static Config config() {
+    var properties = Map.<String, Object>of("maximum-size", 10);
+    return ConfigFactory.parseMap(properties)
+        .withFallback(ConfigFactory.load().getConfig("caffeine.simulator"));
+  }
+}


ben-manes:
I think I had this flagged by Claude as well when after I had it port Lirs v2 from the C++ repository. When I hand wrote v1, I was careful to try to match the C reference and was 0.01% off. Claude found and fixed that final mistake and verified it was a match on all traces. I believe it raised your issue here due to the Lirs2 comment, but determined it was in the Lirs v1 parser rather than the policy itself. Either way, I wanted to match the [C reference](https://github.com/ben-manes/caffeine/tree/master/simulator/src/main/resources/com/github/benmanes/caffeine/cache/simulator/parser/lirs) to honor the authors' intent. When writing v1 and manually tracing, I had found and mimic'd the correlated references behavior. Anyway, I think any changes need to match the v1/v2 reference code which this either has no effect or diverges.
 
<details><summary>Email with Dr. Song</summary>

> Sent: Wednesday, September 2, 2015 2:01 AM
> To: Song Jiang
> Cc: Xiaodong Zhang; sjiang
> Subject: Re: LIRS
>  
> I found the point where the two diverged in multi1. After warmup, if a HIR block is accessed twice in a row then your version will not promote it to a LIR block. I > added a check so a HIR to LIR promotion will only occur if the block is in the stack and is not at the top already. With that the change multi1 miss rate matched. Can > you explain why this is needed and where in the paper this is described?
> 
> ==> This is just a small tweak in the program to avoid counting "correlated references", similar to the thing in your plan to handle multiple references (to the same > block) too close to be considered as different ones. The paper about 2Q discusses this issue: http://www.vldb.org/conf/1994/P439.PDF
> 
> ==> This is not described in the paper.
> 
</details>

I asked Claude but the session memory was lost, so here is what is came up with and recommended against this change. Either way, we'd want to verify across traces that the Java version matches the C/C++ exactly.

<details><summary>Claude's summary</summary>

> Verdict: your recollection is right — decline the PR. But one line in your draft is factually wrong and a sharp reader would rebut it instantly.
> 
> What the sessions/history show
> 
> The decision was deliberate and per-policy, not an oversight:
> - 93b741925 (May 16) — the "0.01% off" fix you remember (onHirWarmupMiss now puts the node on S). Its message: "matches both C references bit-for-bit on loop, multi1/2/3, 2_pools, cpp, cs, scan, DS1, and
> S3."
> - dcd450bc6 (May 17) — "Skip consecutive duplicate accesses in ClockPro and ClockProPlus". It does not touch LirsPolicy, and explicitly excludes ClockProSimple ("its coarser state machine regresses
> slightly").
> - 067f92dbb — Lirs2 born with the guard.
> 
> So LirsPolicy was never given the guard, on purpose. (The session that reasoned it out is past the 30-day retention — the clockpro-dedup-origin memory's origin session is gone — but the commits + comments
> preserve it.)
> 
> The mechanism (this is the defensible reason, and it's empirical)
> 
> Both references collapse the dup with if (ref == last_ref) continue; — but they increment the hit-rate denominator before that line (warm_pg_refs++ in Song's lirs.c; Zhong reports mPageFaultNum /
> mTraceLength, and mTraceLength still contains the dup). So a consecutive dup stays in the denominator as a guaranteed non-miss — i.e. a hit. The continue removes it from the algorithm, not from the rate.
> 
> Current LirsPolicy reproduces that exactly: after any access the block sits at the top of stack S, so a consecutive re-access is a no-op on S/Q state, and it records that second access as a hit. Misses
> match, total refs match, so the rate matches — which is why 93b741925 is bit-for-bit on cs, and cs has 101 consecutive dups (I measured the whole canonical set: multi1=32, multi2=71, multi3=103, cpp=14,
> gli=41, ps=45, sprite=3952; loop/scan/zigzag/2_pools=0).
> 
> The PR puts its guard above recordOperation(), so a dup records nothing and leaves the denominator. That changes the rate on every dup-bearing trace and breaks the bit-for-bit match you validated. Its own
> headline test proves it: record(1); record(1) → it asserts hitCount == 0, but the reference yields a 50% hit rate there (1 miss / 2 refs) = current code's 1 hit. The flagship test encodes a divergence
> from Song's accounting.
> 
> So your instinct — "this either has no effect or diverges" — is precisely correct: no-op on the 4 dup-free traces, divergence on the 8 with dups.
> 
> Why the siblings dedup but v1 must not
> 
> - Lirs2 — the instance role-swap is genuinely non-idempotent. The very comment the PR cites says so: *"not idempotent (unlike LIRS)"* — that parenthetical is the whole point, and the PR author read past
> it.
> - ClockPro/Plus — reference bit is idempotent on a hit, but the adaptive coldTarget would be adjusted twice.
> - ClockProSimple / LIRS v1 — deliberately omitted (Simple regresses; v1 is idempotent and already matches).
> 
> The PR's "siblings have it, so LIRS should too" is a false inference, and its "reorders S / demotes a different LIR block" claim is simply untrue for v1 (the block is at the top of S after every access,
> so the second access touches nothing).
> 
> Suggested reply (your voice, mechanism corrected)
> 
> ▎ Thanks for digging into this — Claude actually flagged the same thing when I ported Lirs2, and we traced it down then.
> ▎
> ▎ The dedup is intentional and honored, but it belongs only where a re-access isn't idempotent. The reference (Song's lirs.c, and Chen Zhong's C++ that Lirs2 is ported from) collapses correlated
> ▎ references with if (ref == last_ref) continue; — but it has already incremented the hit-rate denominator (warm_pg_refs; mTraceLength in Zhong's) before that line. So a consecutive duplicate stays in the
> ▎ denominator as a non-miss, i.e. a hit. record() here reproduces that exactly: after any access the block is at the top of stack S, so a repeated key is a no-op on the stack/queue and is recorded as a
> ▎ hit. That's why LirsPolicy matches both references bit-for-bit on the canonical traces, including cs, which has 101 consecutive duplicates.
> ▎
> ▎ This guard returns before recordOperation(), which drops the duplicate from the denominator entirely — so it changes the hit rate on every trace with back-to-back dups (cs, multi1/2/3, cpp, gli, ps,
> ▎ sprite) and breaks that match. The test makes it concrete: record(1); record(1) asserts hitCount == 0, but the reference reports a 50% rate there (one miss over two references). So the change diverges
> ▎ from the reference rather than matching it.
> ▎
> ▎ Lirs2 and ClockPro need the guard because their per-access transitions aren't idempotent — Lirs2's instance role-swap (the comment you quote says "not idempotent (unlike LIRS)"), and ClockPro's adaptive
> ▎ coldTarget. Plain LIRS has neither, so the guard is a no-op at best and a divergence at worst. I'd rather keep it faithful to the authors' reference. Appreciate the careful read though!
</details>

ben-manes:
Here's a comparison run. Seems like we can close this?

```
⏺ Verified. Here's the empirical result — I ran Song's lirs.c directly (compiled, on cs at the three canonical sizes) and LirsPolicy both ways:

  ┌──────────┬──────────────────────────┬────────────────────────┬────────────────────────┐
  │ cs, size │   C reference (lirs.c)   │ LirsPolicy today (raw) │ with PR guard (dedup)  │
  ├──────────┼──────────────────────────┼────────────────────────┼────────────────────────┤
  │ 500      │ 4717 miss / 6781 → 30.4% │ 4717 / 6781 → 30.44% ✓ │ 4717 / 6680 → 29.39% ✗ │
  ├──────────┼──────────────────────────┼────────────────────────┼────────────────────────┤
  │ 1000     │ 2744 miss / 6781 → 59.5% │ 2744 / 6781 → 59.53% ✓ │ 2744 / 6680 → 58.92% ✗ │
  ├──────────┼──────────────────────────┼────────────────────────┼────────────────────────┤
  │ 2000     │ 1409 miss / 6781 → 79.2% │ 1409 / 6781 → 79.22% ✓ │ 1409 / 6680 → 78.91% ✗ │
  └──────────┴──────────────────────────┴────────────────────────┴────────────────────────┘

  What this shows, concretely:

  - Current LirsPolicy is bit-for-bit with Song's reference — identical miss counts (4717/2744/1409) over the identical denominator (6781, which includes the 101 dups as warm_pg_refs).
  - The PR keeps the misses identical but drops the denominator to 6680 — so it changes nothing about what's cached; the dups current counts as hits (exactly 101: hits go 2064→1963, 4037→3936, 5372→5271)
  are simply removed from the rate. The result diverges from the reference by 0.3–1.05 points on cs alone. So it breaks the match in 93b741925, opposite of its stated goal.

  (The "PR (dedup)" column is exact, not an approximation: the PR's guard returns before recordOperation(), so its only effect is to suppress the record() call on a consecutive dup — identical to feeding
  the deduped stream. Happy to apply the literal 6-line patch and re-run if you want that in the thread too.)

  On your clarification — yes, you're right

  The dedup lives in the C reference's trace-reading loop (LIRS_Repl in lirs.c, if (ref==last_ref) continue;), and your Java LirsTraceReader correctly does not dedup. I'd misread your "v1 parser" as the
  Java reader — my mistake. The one nuance worth carrying into the reply: the C reader increments warm_pg_refs (the denominator) before that continue, so its dedup removes the dup from the algorithm, not
  from the rate — the dup still scores as a non-miss. That's why a Java reader-level dedup would be wrong too (it'd drop dups from the denominator, same defect as the PR). The port matches the reference
  because the policy counts the dup as a hit, which the table now confirms. So phrase it as "the C reference's trace reader" rather than just "parser," so nobody checks LirsTraceReader.java and finds
  nothing.

  Reply, with the numbers folded in

  ▎ Thanks for the careful read — Claude flagged this same thing when I ported Lirs2, and we worked through it then.
  ▎
  ▎ The dedup is intentional and honored, but it belongs only where a re-access isn't idempotent. Song's reference dedups in its trace-reading loop (if (ref == last_ref) continue; in LIRS_Repl) — but it has
  ▎ already incremented the hit-rate denominator (warm_pg_refs) before that line, so a consecutive duplicate stays in the denominator as a guaranteed non-miss, i.e. a hit. The dedup drops the duplicate
  ▎ from the algorithm, not from the rate.
  ▎
  ▎ LirsPolicy.record() reproduces that exactly: after any access the block sits at the top of stack S, so a repeated key is a no-op on the stack/queue and is simply recorded as a hit. That's why it matches
  ▎ the reference bit-for-bit — I just re-ran Song's lirs.c on cs to confirm:
  ▎
  ▎ | size | lirs.c            | LirsPolicy | with this PR        |
  ▎ |------|-------------------|------------|---------------------|
  ▎ | 500  | 4717/6781 = 30.4% | 30.4% ✓    | 4717/6680 = 29.4% ✗ |
  ▎ | 1000 | 2744/6781 = 59.5% | 59.5% ✓    | 58.9% ✗             |
  ▎ | 2000 | 1409/6781 = 79.2% | 79.2% ✓    | 78.9% ✗             |
  ▎
  ▎ The guard returns before recordOperation(), so it removes the 101 consecutive duplicates in cs from the denominator and shifts the hit rate away from the reference — same eviction behavior, different
  ▎ (and now non-matching) accounting. So this diverges from the reference rather than matching it.
  ▎
  ▎ Lirs2 and ClockPro carry the guard because their per-access transitions aren't idempotent — Lirs2's instance role-swap (the comment you quote even says "not idempotent (unlike LIRS)") and ClockPro's
  ▎ adaptive coldTarget. Plain LIRS has neither, so the guard is a no-op on dup-free traces and a divergence on the rest. I'd rather keep it faithful to the authors' reference. Really appreciate the
  ▎ contribution though — please keep them coming!
```

sahvx655-wq:
You're right, and I had the mechanism backwards. I went back through record() and traced it on cs: every path leaves the accessed block at the top of stack S, so a back-to-back repeat is a genuine no-op on S and Q, and the existing code already scores that second access as a hit. That's exactly the reference's accounting, since warm_pg_refs is incremented before the `if (ref == last_ref) continue`, so the duplicate stays in the denominator as a non-miss. My "reorders S / demotes a different LIR block" claim just doesn't hold for v1, because the second access never touches the stack.

So the guard returning before recordOperation() drops the dup from the denominator, which pulls the rate away from lirs.c on every trace with back-to-back dups (cs, multi1/2/3, and the rest) rather than matching it. The siblings carry the guard because their per-access transitions genuinely aren't idempotent, which plain LIRS isn't. Closing this. Thanks for the careful trace and the numbers.

PR #1985, opened 2026-06-25
remove the doubled prefix from sampled policy names

sahvx655-wq:
Running the simulator over the bundled traces, every sampled policy comes back in the results table as `sampled.sampled.Lru`, `sampled.sampled.Mru` and so on, with a doubled prefix. `SampledPolicy.EvictionPolicy.label()` already returns `sampled.Lru`, but the constructor prepends another `sampled.` on top of it, so the sampled family is the only one whose reported name disagrees with the name it is registered and configured under. `LinkedPolicy` and `FrequentlyUsedPolicy` pass `policy.label()` straight through to `PolicyStats`.

Drop the redundant prefix so the reported name matches the registered name. That label is what the table and CSV report sort and join on, so a `sampled.Lru` row currently cannot be matched back to the `sampled.Lru` requested in the config, and any tooling correlating results to the configured policy quietly misses the sampled entries. Hit rates and every other figure are unchanged; only the label differs.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sampled/SampledPolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sampled/SampledPolicy.java
@@ -69,7 +69,7 @@ public final class SampledPolicy implements KeyOnlyPolicy {
   long tick;
 
   public SampledPolicy(Admission admission, EvictionPolicy policy, Config config) {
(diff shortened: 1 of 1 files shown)


ben-manes (approved):
oh nice find. I somehow never noticed this despite all these years of scanning the report when tuning the policy.

sahvx655-wq:
It only surfaced because I was joining the report's policy column back to the config to line up which run was which, and the sampled rows wouldn't match on the name. Easy thing to scan past for years once you already know which row is which, the label just never lines up with anything that reads it back.

PR #1987, opened 2026-07-04
count the eviction into dclock's refault distance

sahvx655-wq:
DClock stamps each evicted shadow with the nonresident age (evictions + activations) and later reads it back as the refault distance that gates optimistic activation. The stamp is taken before the eviction is counted, so every recorded distance sits one slot above the cited linux/mm/workingset.c reference, which advances the age through workingset_age_nonresident ahead of the read that stamps the page. A shadow that refaults at exactly the active-list size is then demoted to inactive at the boundary the kernel would promote on.

Counting the eviction before the stamp lines the distance up with the reference and restores that boundary promotion. It is not purely cosmetic: the page that should have been pinned in the active list stays inactive, is caught by the next scan, and misses on its next reuse, so DClock's hit rate drifts under the algorithm it models. The regression drives a two-entry cache where the hot key only survives a following scan if its boundary refault activated it.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/DClockPolicy.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/DClockPolicy.java
@@ -205,8 +205,10 @@ public final class DClockPolicy implements KeyOnlyPolicy {
       Node victim = requireNonNull(headInactive.prev);
       policyStats.recordEviction();
 
-      victim.nonResidentAge = currentNonResidentAge();
+      // Advance the age before stamping so the shadow's refault distance counts this eviction,
+      // matching workingset_age_nonresident running ahead of the read in the kernel reference.
       evictions++;
+      victim.nonResidentAge = currentNonResidentAge();
       inactiveSize--;
       victim.remove();
       victim.status = Status.NON_RESIDENT;
--- a/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/DClockRefaultDistanceTest.java
+++ b/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/irr/DClockRefaultDistanceTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2026 Ben Manes. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.benmanes.caffeine.cache.simulator.policy.irr;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.Map;
+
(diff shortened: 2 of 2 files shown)


ben-manes:
Claude didn’t seem to agree. Can you ask your agent to double check? We mirror Linux as the only implementation and description of this policy.

——

I’ve verified this against the actual kernel source. The PR is not correct — its premise about the Linux implementation is backwards, and the current Caffeine code already matches the kernel.

### What the PR claims

The PR says the kernel “advances the age through workingset_age_nonresident ahead of the read that stamps the page,” and therefore the stamp should count the eviction itself (i.e., evictions++ should happen before stamping victim.nonResidentAge).

### What the kernel actually does

In workingset_eviction() the order is the opposite of what the PR claims — the age is read into the stamp first, then advanced:

```c
eviction = atomic_long_read(&lruvec->nonresident_age);   // ← read / stamp
eviction >>= bucket_order[file];
workingset_age_nonresident(lruvec, folio_nr_pages(folio)); // ← advance, AFTER the read
return pack_shadow(memcgid, pgdat, eviction, ...);
```

So the shadow is stamped with the age excluding its own eviction. The +nr_pages advance lands only on subsequent evictions/activations.

### What Caffeine currently does

```c
victim.nonResidentAge = currentNonResidentAge();  // read/stamp  ≡ eviction = atomic_long_read(...)
evictions++;                                       // advance     ≡ workingset_age_nonresident(...)
```

This mirrors the kernel exactly: stamp before the counter advances. The PR’s reordering (evictions++ before the stamp) would make Caffeine stamp one slot higher than the kernel, so every recorded refault distance would come out one smaller than the reference.

### The boundary case is already handled

The kernel activates when refault_distance <= workingset_size (its check is if (refault_distance > workingset_size) goto out). Caffeine’s onNonResidentHit uses refaultDistance(node) <= activeSize — same boundary, inclusive. A page refaulting at exactly the active size is already promoted by the current code. The PR’s stated failure (“demoted to inactive at the boundary the kernel would promote on”) doesn’t exist; the off-by-one it describes is the one its own change would introduce. After the change, Caffeine would activate pages the kernel rejects (kernel-equivalent distance activeSize + 1).

### Verdict

Not a valid flaw in our code. The current implementation is faithful to mm/workingset.c; the PR inverts the read/advance ordering and would diverge from the reference by one slot in the wrong direction. The included test only “passes” because it’s constructed around the shifted distances, not because it reflects kernel behavior.

ben-manes:
Oh I asked it to check the exact commit hash in our javadoc vs master, and it’s a kernel change. The hash is outdated

ben-manes:
https://github.com/torvalds/linux/commit/aeddcee6c17bd8cf80675495d39c4daceaf5b506.patch

sahvx655-wq:
You're right, and the commit you linked settles it against me. I pulled aeddcee6c17b ("mm: workingset: clarify eviction order and distance calculation") and it moves workingset_age_nonresident from before the atomic_long_read to after it. The message is explicit that the old advance-before-read ordering left things "off by one, so the real number of slots would be (R - E) + 1", and that reorder is what corrected it. Stamping after evictions++ (my patch) is that same pre-fix ordering, so it makes the recorded distance one smaller and over-activates at the boundary rather than fixing anything. Master reads then advances, and the existing code (stamp, then evictions++) already matches that, so there's no bug here to fix.

What threw me was the javadoc hash. It pins ac3a0c8472, which predates the reorder, so at the version we cite the kernel genuinely did advance first and my change agreed with it. But that snapshot still carries the off-by-one the kernel later removed, so it's the stale reference that's wrong, not the code. Happy to close this. The only real follow-up, if you think it's worth it, is bumping the workingset.c hash in the DClockPolicy javadoc to a current commit so nobody else reads the old ordering as canonical.

ben-manes:
I folded the javadoc updates in with another commit. thanks for pointing it out.

PR #1990, opened 2026-07-10
flush the console report writer instead of closing System.out

sahvx655-wq:
The console report path in `TextReporter.makeWriter` hands `print()` a `PrintWriter` wrapping `System.out`, and `print()` closes that writer in its try-with-resources. Closing the writer closes `System.out` with it, so once a report has been rendered the process has no stdout left: the "Executed in ..." line that `Simulator.main` prints next lands on a closed stream and is silently swallowed, with `System.out.checkError()` flipping to true right after the report. Every default console run hits this, and anything that later writes to stdout would disappear the same way.

The reporter does not own `System.out`, so the console writer now flushes on close instead of propagating the close, leaving the stream open. Keeping that behaviour in `makeWriter` means the file-output branch still closes its writer as before and `print()` stays a generic try-with-resources over any `Writer`. Added a regression test that swaps in a close-tracking stdout and checks the reporter leaves it open.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/report/TextReporter.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/report/TextReporter.java
@@ -20,6 +20,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Locale.US;
 import static java.util.Objects.requireNonNull;
 
+import java.io.FilterWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.UncheckedIOException;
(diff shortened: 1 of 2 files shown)


ben-manes:
I noticed the output needs a newline or else the "Executed in..." falls on the table's border. Maybe a `console.println()` before the `flush()` when closing?

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/report/TextReporter.java
```diff
@@ -93,7 +94,14 @@ private ImmutableSet<String> getHeaders(Collection<PolicyStats> results) {
   private Writer makeWriter() throws IOException {
     String output = settings.report().output();
     if (output.equalsIgnoreCase("console")) {
-      return new PrintWriter(System.out, /* autoFlush= */ true, UTF_8);
+      // print() closes the writer, but System.out belongs to the process, not the reporter.
+      // Flush instead of closing so the caller's later output is not dropped onto a dead stream.
```
ben-manes: I think you can probably drop this comment. I know AI likes to drop lots of notes in the code based on the current session, but not its not the most important thing to callout to a human reading the code. When useful enough I'll ask it to update the .claude/docs so its available to AI. Here I think merely the git blame is enough to let anyone interested to understand your bug fix.

sahvx655-wq:
Good catch, close() now does a println before the flush. Checked it with a quick synthetic LRU run and the "Executed in..." line sits on its own line under the bottom border now instead of running into it.

simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/report/TextReporter.java
sahvx655-wq: Fair enough, dropped it. The blame is enough to find the why.

ben-manes (approved):

PR #1991, opened 2026-07-12
gate the minisim window resize on a full cache

sahvx655-wq:
The minisim strategy for sketch.HillClimberWindowTinyLfu resizes the admission window at every period boundary without checking whether the cache has filled, unlike its siblings: AbstractClimber.adapt holds while the cache is under capacity and IndicatorClimber only accumulates its sample once full. A trace whose working set stays below the window capacity therefore keeps the window under-filled, yet at the boundary decreaseWindow still sizes its move count from the window capacity (quota = min(amount, maxWindow)) rather than the resident node count. The move loop walks off the window list onto the headWindow sentinel and the run dies with a NullPointerException in Node.appendToHead, with windowSize going negative and tripping the checkState right after; I traced it back from that stack after replaying a small-working-set trace.

The fix holds the adaptation while the cache is not full, mirroring AbstractClimber, so a resize only fires once window occupancy equals the window capacity and the move count can no longer exceed the resident nodes. Keeping the guard in the climber leaves the policy's resize maths untouched and puts minisim in the same full-cache-only regime the other strategies already rely on. Left unfixed this aborts the whole simulation for any minisim run over a trace with few distinct keys, so the added test replays three keys past a shortened period and reproduces the crash without the change.

### DIFF
--- a/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/sim/MiniSimClimber.java
+++ b/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/sim/MiniSimClimber.java
@@ -89,7 +89,7 @@ public final class MiniSimClimber implements HillClimber {
   @Override
   public Adaptation adapt(double windowSize,
       double probationSize, double protectedSize, boolean isFull) {
-    if (sample <= period) {
+    if (!isFull || (sample <= period)) {
       return Adaptation.hold();
     }
 
--- a/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/MiniSimUnfilledResizeTest.java
+++ b/simulator/src/test/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/MiniSimUnfilledResizeTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2026 Ben Manes. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.benmanes.caffeine.cache.simulator.policy.sketch.climbing;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+import com.github.benmanes.caffeine.cache.simulator.policy.AccessEvent;
+import com.github.benmanes.caffeine.cache.simulator.policy.Policy;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * The minisim climber adapts the admission window at its period boundary regardless of whether the
+ * cache is full. A trace whose working set stays below the window capacity keeps the window
+ * under-filled, so the period-boundary resize walks past the resident window nodes.
+ *
+ * @author sahvx655@gmail.com (Sahana Bogar)
+ */
+final class MiniSimUnfilledResizeTest {
+
+  @Test
+  void unfilledWindowResize() {
+    var policy = policy();
+    // Three keys against a 1,000-entry cache leaves the ~1% window far below its capacity; replay
+    // past the (shortened) minisim period so the boundary adaptation fires while under-filled.
+    for (int i = 0; i < 200; i++) {
+      policy.record(AccessEvent.forKey(i % 3));
+    }
+    assertThat(policy.stats().requestCount()).isEqualTo(200L);
+  }
+
+  private static Policy policy() {
+    var config = ConfigFactory.parseString("""
+        maximum-size = 1000
+        hill-climber-window-tiny-lfu {
+          strategy = [ minisim ]
+          percent-main = [ 0.99 ]
+          minisim.period = 100
+        }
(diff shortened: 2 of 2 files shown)


ben-manes:
Thanks! I had claude verify this didn't cause any surprise regressions using Lirs' loop when starting in an LRU setting and Corda when starting in an MRU/LFU setting. lgtm.

Reviewing your comments wrt the `IndicatorClimber` which seems to adapt early too, right? It has a hardcoded sample period, which also looks a bit odd given the variable cache sizes between traces and our config uses to allow experimenting with different tuning values. Should the `isFull` guards be placed in [HillClimberWindowTinyLfuPolicy](https://github.com/ben-manes/caffeine/blob/c3e027395f65f96065dbf1f46ae4b4d06ce29c99/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/HillClimberWindowTinyLfuPolicy.java#L210-L225) and not passed down to the climbers to handle? It looks like I added it to minimally modify the existing climbers when adding a new [type](https://github.com/ben-manes/caffeine/commit/2b2381b256ae1583278c5ca7bc1db1d90d07173e) (gradient descent) and likely to try to avoid similar crashes.

https://github.com/ben-manes/caffeine/blob/c3e027395f65f96065dbf1f46ae4b4d06ce29c99/simulator/src/main/java/com/github/benmanes/caffeine/cache/simulator/policy/sketch/climbing/inference/IndicatorClimber.java#L56-L70

---
I tried asking Claude these questions, which seemed to confirm that the minisim and indicator are safe from crashes but also too brittle. That's out of scope here but something we should be aware of, either just to note and backlog or resolve later on.

<details>

Confirmed all three. Commit 2b2381b256 (Dec 2018, "Gradient descent optimizers") is exactly where isFull entered the HillClimber interface and the per-climber guards appeared — your recollection is right. Here are answers to the three questions.

#### 1. Does IndicatorClimber adapt early? No — but only by luck of an implicit guard

Its adapt() has no explicit !isFull check, so on a quick read it looks as exposed as minisim was. It isn't, because onHit/onMiss gate the feed on isFull:
```java
public void onHit(long key, QueueType queue, boolean isFull) {
  if (isFull) { indicator.record(key); }   // record() does sample++
}
```
Indicator.getSample() only advances inside record(), so the == 50_000 trip can't happen until 50,000 post-fill accesses have elapsed. And once isFull, the window is held exactly full (windowSize == maxWindow is invariant when full — the window fills before main and stays pinned), so its decreaseWindow move count can never exceed the resident nodes. So it can't adapt early and can't hit the crash.

The clean way to state the asymmetry in your reply: minisim is the only climber that gates neither its sample counter nor its adaptation on isFull — onAccess does sample++ unconditionally and adapt fired on sample > period. IndicatorClimber gates the counter (which implicitly gates adaptation); the AbstractClimber family gates both explicitly. minisim gated nothing, so it was uniquely broken.

#### 2. The hardcoded 50_000 — yes, it's the outlier, on two axes

Every other climber's sample period is configurable, and most scale with cache size:
```
┌────────────────────────┬───────────────────────────────┬──────────────────────┬───────────────────┐
│        Climber         │         Sample period         │    Configurable?     │ Scales with size? │
├────────────────────────┼───────────────────────────────┼──────────────────────┼───────────────────┤
│ hill / gradient family │ percentSample() * maximumSize │ yes (percent-sample) │ yes               │
├────────────────────────┼───────────────────────────────┼──────────────────────┼───────────────────┤
│ minisim                │ minisim.period (1e6)          │ yes                  │ no (fixed count)  │
├────────────────────────┼───────────────────────────────┼──────────────────────┼───────────────────┤
│ indicator              │ 50_000 literal                │ no                   │ no                │
└────────────────────────┴───────────────────────────────┴──────────────────────┴───────────────────┘
```
The indicator { k, ss-size } block is the natural home for a sample-size key. One nuance worth conceding in the reply: a fixed count is actually defensible here — the indicator estimates workload skew (top-k over a stream-summary), which is a property of the access distribution, not the cache, so you want a stable sample size regardless of maximumSize rather than one that scales. The real gap is that it's baked in and un-tunable while everything around it is a knob — so it can't be swept, and on a short trace / late-filling cache it may silently never reach 50k full accesses and never adapt at all (same failure-to-adapt mode minisim has with its 1e6 default).

#### 3. Should the guard live in climb() instead of each climber? Yes

This is the right instinct, and it directly fixes the root cause: the guard is a cross-cutting invariant of the policy ("don't resize the window until it can be full"), not something each climber should re-derive. Leaving it per-climber is what let minisim slip through. Two levels:

Minimal safety net (what I'd do) — gate only the adapt dispatch in climb():
```java
Adaptation adaptation = isFull
    ? climber.adapt(windowSize, probationSize, protectedSize, isFull)
    : Adaptation.hold();
```
Makes the crash structurally impossible for every climber, present and future, with zero behavior change — each climber that would have adapted while !isFull already returned hold() (and for minisim this is the fix). It subsumes PR #1991's one-liner: with this in place the per-climber change is unnecessary. isFull still needs to reach onHit/onMiss for sample gating, so it stays in the interface.

Full centralization (bigger, needs re-validation) — early-return in climb() before recording or adapting, and drop isFull from the HillClimber interface entirely:
```java
private void climb(long key, @Nullable QueueType queue, boolean isFull) {
  if (!isFull) { return; }
  ...
}
```
Behavior-identical for the AbstractClimber family and IndicatorClimber (they already no-op while !isFull), but it changes minisim: its minis currently record from the very first access, and this would restrict them to post-fill accesses. Arguably more consistent, but it perturbs the estimator's warmup, so it'd want a re-run on your loop/corda checks before trusting it.

My recommendation: take the minimal central guard as the actual fix (it retires this whole bug class and makes the per-climber !isFull checks redundant belt-and-suspenders), and treat "remove isFull from the interface + gate minisim's sampling" as a separate cleanup only if you're happy to re-validate minisim.

One caveat if you centralize and then remove the per-climber guards: AbstractClimber.adapt also does checkState(sampleSize > 0) right after its !isFull return — that assertion (and the sample-reset bookkeeping) would then run pre-fill unless the central guard short-circuits before the call, so the early-return form is the one to use, not just replacing the dispatch.

</details>

ben-manes (approved):

PR #1992, opened 2026-07-14
handle an expired prior in postProcess read and update actions

sahvx655-wq:
Cache.invoke reads the entry, runs the user's EntryProcessor, then passes the result to postProcess, which re-reads the clock and nulls its local expirable when the prior entry has lapsed since that initial read. The NONE, CREATED/LOADED and DELETED branches all cope with that null, but READ and UPDATED assert requireNonNull(expirable), so an entry that expires lazily mid-processing throws NullPointerException back through invoke as an EntryProcessorException. On the UPDATED path recordPuts and CacheWriter.write both run before the assertion, so a write-through fires and the put statistic ticks up on an operation that then fails and stores nothing, leaving the backing store and the cache out of step.

Give both branches the handling their siblings already have: publish the prior's deferred expiration and, for an update, finish it as a creation since an update whose prior has gone is really a create, honouring a zero creation expiry the way the CREATED branch does. Keeping it in postProcess means the per-key compute still holds the entry, so the expiry decision stays atomic with the write. The existing postProcess tests drive the NONE, CREATED and DELETED expired-prior cases directly, so I added the read and update counterparts, which throw before this change and pass after.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
@@ -901,7 +901,12 @@ public class CacheProxy<K, V> implements Cache<K, V> {
         publishExpiredPrior(entry.getKey(), expiredValue);
         return expirable;
       case READ: {
-        setAccessExpireTime(entry.getKey(), requireNonNull(expirable), 0L);
+        if (expirable == null) {
+          // the read entry expired lazily during processing, so publish its expiration and drop it
+          publishExpiredPrior(entry.getKey(), expiredValue);
+          return null;
+        }
+        setAccessExpireTime(entry.getKey(), expirable, 0L);
         return expirable;
       }
       case CREATED:
@@ -922,9 +927,22 @@ public class CacheProxy<K, V> implements Cache<K, V> {
         return new Expirable<>(value, expireTimeMillis);
       }
       case UPDATED: {
+        if (expirable == null) {
+          // the prior expired lazily during processing, so the write is a creation, not an update
+          publishToCacheWriter(writer::write, () -> entry);
+          V created = copyOf(requireNonNull(entry.getValue(), "Expected a new value but was null"));
+          publishExpiredPrior(entry.getKey(), expiredValue);
+          long expireTimeMillis = getWriteExpireTimeMillis(/* created= */ true);
+          if (expireTimeMillis == 0) {
+            dispatcher.publishExpired(this, entry.getKey(), created);
+            return null;
+          }
+          statistics.recordPuts(1L);
+          dispatcher.publishCreated(this, entry.getKey(), created);
+          return new Expirable<>(created, expireTimeMillis);
+        }
         publishToCacheWriter(writer::write, () -> entry);
         statistics.recordPuts(1L);
-        requireNonNull(expirable, "Expected a previous value but was null");
         V value = copyOf(requireNonNull(entry.getValue(), "Expected a new value but was null"));
         dispatcher.publishUpdated(this, entry.getKey(), expirable.get(), value);
         @Var long expireTimeMillis = getWriteExpireTimeMillis(/* created= */ false);
--- a/jcache/src/test/java/com/github/benmanes/caffeine/jcache/CacheProxyTest.java
+++ b/jcache/src/test/java/com/github/benmanes/caffeine/jcache/CacheProxyTest.java
@@ -772,6 +772,61 @@ final class CacheProxyTest {
   }
 
   @Test
+  void postProcess_read_expiredPrior_publishesExpiredForPrior() {
+    try (var fixture = jcacheFixture(Mockito.mock(), Mockito.mock(), Mockito.mock())) {
+      var expired = new Expirable<>(VALUE_1, fixture.currentTime().toMillis() - 1);
+      EntryProcessorEntry<Integer, Integer> entry = Mockito.mock();
+      when(entry.getAction()).thenReturn(Action.READ);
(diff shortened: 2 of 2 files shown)


ben-manes:
Thanks for finding this!

It appears to be a regression from https://github.com/ben-manes/caffeine/commit/6e97733f2bc64fc3c69c00d756ea8911541ecb02, an AI audit driven change that I made recently. I think that was wrong and should be reverted, with tests to assert/pin the behavior.

My reasoning is that when the `invoke` began the entry was valid and the entry is being operated on atomically under a `Map.compute` lock. If the operation lengthened the entry's expiration time then it should become valid. If it deleted it then that takes precedence over expiration, as technically happened first. If they updated it then that could be derived from the read, so it should be treated like a replacement and not creation. The `NONE` case was when nothing happened at all, so the expiration could be done proactively as an optimization. It could also be simply ignored and defer to the cache's maintenance to expunge expired entries. The only valid case was NONE and I mistakenly allowed AI to change that.

My plan is to restore the old behavior, add tests to pin it, and then close this. Thank you for flagging this mistake pre-release. Luckily jcache was largely a failed standard so it has low adoption, so the regression's impact would have been minimal.

ben-manes:
hah, and AI disagrees with both of us with some reasonable arguments. I will need to look at it more and harden it up regardless.

> I've now traced the full path — the invoke() read, the postProcess timestamp flow, the EntryProcessorEntry state machine, and both test files. Here's my honest read, > and it cuts in both directions.
> 
> Bottom line
> 
> You're right to decline #1992 — but not for the reason in your message, and I don't think 6e97733 was wrong for the cases it actually affects. Reverting to "only NONE > expires" would reintroduce a real bug. Let me show why.
> 
> The premise to correct: postProcess never reaps an entry that was valid at the start
> 
> The thing your message worries about — an entry valid at the start of invoke() getting expired at the end — cannot happen, and it has nothing to do with the fix. The > timestamp is frozen at the initial read:
> 
> - invoke() reads the entry and, for a present non-eternal entry, assigns millis = currentTimeMillis() right there in the hit/miss test (CacheProxy.java:853-860).
> - postProcess receives that same millis and reuses it (:890-898). It only re-reads the clock through the currentTimeMillis == 0 sentinel — which doesn't fire when > millis was already stamped.
> 
> So hasExpired(millis) in postProcess returns the identical verdict it returned at the initial read. An entry that was valid at the start is valid at the end — already > "resurrected," by construction. The branch fires on exactly one thing: entries that were already expired when invoke() started, where the processor correctly saw exists> () == false and JCache owes a lazy EXPIRED event. Your "if valid at start, keep it" rule is already the behavior.
> 
> Why #1992 is chasing a phantom
> 
> READ and UPDATED are unreachable alongside an expired prior. Straight from the state machine (EntryProcessorEntry.java), where hasEntry = (value != null) is fixed at > construction — and value is null whenever the entry was absent or expired at the initial read:
> 
> - READ is set only in getValue() when value != null from the NONE state (:63-67) → requires hasEntry == true.
> - UPDATED is set only in setValue() when hasEntry is true (:87-90); otherwise you get CREATED.
> 
> So READ/UPDATED ⟹ hasEntry == true ⟹ the prior was present and unexpired at the initial read ⟹ millis stamped non-zero ⟹ postProcess never nulls expirable. The > requireNonNull(expirable) on those two branches is an always-true invariant, not a live NPE.
> 
> The PR's tests only go green because they Mockito.mock() the entry and force getAction() to return READ/UPDATED next to an already-expired expirable — a (action, prior) > pair the real EntryProcessorEntry can't emit. Contrast the existing CREATED/DELETED expired-prior tests: same mock style, but those states (hasEntry == false + setValue/> remove) are genuinely reachable. #1992's are not.
> 
> Why 6e97733 itself was not wrong
> 
> The cases it actually changed — NONE/CREATED/LOADED/DELETED with an already-expired prior — are all reachable, and publishing EXPIRED there is correct JCache > lazy-expiration. In particular the DELETED branch previously fired REMOVED + recordRemovals against the stale expired value (:939-942 shape). Reverting to "only NONE > expires" would re-break that and drop the lazy-EXPIRED events on the create/load paths. So the human "only NONE" logic is the thing I'd push back on — it's wrong for > the reachable mutation cases.
> 
> The one real, small thing 6e97733 did introduce
> 
> Hoisting the expirable = null above the switch made READ/UPDATED look exposed even though the state machine guarantees they aren't. That non-local invariant is exactly > the bait that produced #1992. There's also a latent smell: millis == 0 doubles as both "not yet read" and "clock is at epoch 0," which is the only razor-thin path (a > ticker sitting at 0ms at the initial read) where the re-read could observe expiry for a hit entry — a sentinel collision, not the "expires mid-processing" story the PR > tells.
> 
> Recommendation
> 
> - Decline #1992. It guards an unreachable state and its tests assert on an impossible input.
> - Keep 6e97733 for the reachable cases; don't revert to only-NONE.
> - If you want to kill this at the root so no future audit re-files it: make the READ/UPDATED invariant local — a one-line comment ("READ/UPDATED ⟹ prior existed and was > unexpired at the initial read, so expirable is non-null"), and/or de-fang the == 0 sentinel so the non-null property is provable without appeal to the state machine. > That's the actual fix — not phantom branches.
> 
> And to own it plainly: the audit that produced this shape reasoned locally ("expirable can be null above the switch, therefore READ/UPDATED can deref null") without > discharging the reachability against EntryProcessorEntry. That's the miss. It's a false positive dressed as a null-safety fix.
> 
> Want me to confirm the unreachability empirically — a test that drives a real EntryProcessorEntry through invoke() with an expired-at-start entry and a read/update > processor, asserting getAction() comes back NONE/CREATED/DELETED and never READ/UPDATED? That would settle it without either of us hand-waving.
>

PR #1993, opened 2026-07-15
record put and remove time only on an actual mutation

sahvx655-wq:
**Average put/remove time is charged to operations that change nothing**

Chasing an inflated `getAveragePutTime` reading, I traced every write and delete path in `CacheProxy` and found the timing is recorded unconditionally while the matching count only moves when the entry is genuinely stored or removed. A `putIfAbsent` on a present key, a value-mismatched `replace`/`remove(K, V)`, an absent-key `getAndReplace`/`getAndRemove`, and an empty `putAll`/`removeAll` each add their duration to the numerator with no put or removal in the denominator, so the reported mean is skewed high. The two-arg `replace(K, V)` and the iterator's `remove()` already gate the timing behind the mutation, and the JSR-107 reference implementation only accumulates the time when a put or removal occurs, which points to the other paths being an oversight rather than a deliberate choice.

Left unfixed, any workload that leans on conditional writes reports a put or remove latency several times higher than the operations that were actually counted, which quietly misleads anyone using the JMX statistics to reason about cache cost. Each timing call is now gated on the same condition as its counter, with a regression covering the no-op put and remove variants. Verified with `:jcache:test` and `:jcache:tckTest`.

### DIFF
--- a/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
+++ b/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java
@@ -332,10 +332,8 @@ public class CacheProxy<K, V> implements Cache<K, V> {
     var result = putNoCopyOrAwait(key, value, /* publishToWriter= */ true);
     dispatcher.awaitSynchronous();
 
-    if (statsEnabled) {
-      if (result.written) {
-        statistics.recordPuts(1);
-      }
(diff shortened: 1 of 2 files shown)


ben-manes:
thanks! I had a bunch of other audit fixes in the v3.dev branch, so I rebased yours onto it and merged them all into master.

PR #1996, opened 2026-07-31
Remove unnecessary dependency on apache.commons.text.

chaoren:
It's a rather heavy dependency to pull in for not very much benefit. Internally, we noticed a build metric regression due to this extra dependency.

### DIFF
--- a/caffeine/build.gradle.kts
+++ b/caffeine/build.gradle.kts
@@ -88,7 +88,6 @@ dependencies {
   javaPoetImplementation(libs.guava)
   javaPoetImplementation(libs.javapoet)
   javaPoetImplementation(libs.jspecify)
-  javaPoetImplementation(libs.commons.text)
   javaPoetImplementation(libs.commons.lang3)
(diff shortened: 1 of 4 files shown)


ben-manes:
can you explain why? I assume its for easier vendor'ing but also commons is known. I believe that I did this since commons avoids some allocations and was doing some other cleanups at the time, but I agree its optional.

caffeine/build.gradle.kts
```diff
@@ -88,7 +88,6 @@ dependencies {
   javaPoetImplementation(libs.guava)
   javaPoetImplementation(libs.javapoet)
   javaPoetImplementation(libs.jspecify)
-  javaPoetImplementation(libs.commons.text)
```
ben-manes, line 91: can you remove it from `gradle/libs.versions.toml` as otherwise unused?

ben-manes (dismissed):

caffeine/build.gradle.kts
chaoren: Done.

PR #2003, opened 2026-08-21
Fix upcoming warnings from NullAway 0.14.0

msridhar:
These fix some inconsistent type variable upper bounds in overrides, which will be reported as errors in the soon-to-be-released NullAway 0.14.0

### DIFF
--- a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
+++ b/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
@@ -3298,7 +3298,8 @@ abstract class BoundedLocalCache<K, V> extends BLCHeader.DrainStatusRef
    * @return the computed value
    */
   @SuppressWarnings("GuardedByChecker")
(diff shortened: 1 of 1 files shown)


ben-manes (approved):

caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java
```diff
@@ -3298,7 +3298,8 @@ public String toString() {
    * @return the computed value
    */
   @SuppressWarnings("GuardedByChecker")
-  <T> T evictionOrder(boolean hottest, Function<@Nullable V, @Nullable V> transformer,
+  <T extends @Nullable Object> T evictionOrder(boolean hottest,
+      Function<@Nullable V, @Nullable V> transformer,
```
vlsi, line 3302: Just wondering: should it be `Function<? super @Nullable V, ? extends @Nullable V>`?

msridhar: I didn't touch that part of the function signature in this PR. I was just fixing the new NullAway warnings. That could be useful as a more flexible type, though.

PR #2007, opened 2026-09-02
Remove some unnecessary NullAway suppressions

msridhar:
Computed with a new script I wrote that I'll push to NullAway once further tested.


ben-manes (approved):

msridhar:
FYI the script I used to generate this: https://github.com/uber/NullAway/tree/master/scripts/nullaway-suppression-remover

